PR #1, opened 2022-08-05
Add milvus storage minio service

EricStarer:
Signed-off-by: xingzhao <xing.zhao@zilliz.com>

### DIFF
--- a/core/milvus_storage.go
+++ b/core/milvus_storage.go
@@ -1,11 +1,356 @@
 package core
 
-// MilvusStorage is interface to operate milvus data
-type MilvusStorage interface {
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"go.uber.org/zap"
+	"io"
+	"io/ioutil"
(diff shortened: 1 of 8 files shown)


core/milvus_storage.go
```diff
@@ -1,11 +1,268 @@
...
+	// MultiRemove delete @filePaths.
+	MultiRemove(filePaths []string) error
+	// RemoveWithPrefix remove files with same @prefix.
+	RemoveWithPrefix(prefix string) error
 }
 
 // makes sure MinioMilvusMeta implements `MilvusStorage`
 var _ MilvusStorage = (*MinioMilvusMeta)(nil)
 
 type MinioMilvusMeta struct {
```
wayblink, line 58: I think the name is not proper. We operate data here not meta data. How about MinioMilvusStorage?

wayblink (commented):
Generally LGTM with two comments. Pls add UT for milvus_storage.go.

PR #85, opened 2023-02-09
core: concurrency copy

huanghaoyuanhhy:


zhuwenxing:
```
bulk insert fail, info: failed to seal segment, shard id 1, segment id 439338129502384914, channel 'by-dev-rootcoord-dml_145_439338129502384839v1', error: All attempts results:\nattempt #1:failed to save import segment, reason = failed to get segment 439338129502384914\
```
This error was not met before, so I think we should spend some time to find out the reason, though it was not 100% appearances

wayblink (approved):
LGTM

PR #115, opened 2023-03-29
common: no longer reuse worker

### LINKED ISSUE #117: [Bug]: When the job returns an error, the corresponding goroutine will exit.
### Current Behavior

https://github.com/zilliztech/milvus-backup/blob/25c71c70d55feddb2e7219b8997b65a70d766e6e/internal/common/workerpool.go#L41-L55

When the job returns an error, the corresponding goroutine will exit, resulting in a decrease in consumers and eventually a backlog of tasks.

### Expected Behavior

The number of consumers should not decrease

### Steps To Reproduce

_No response_

### Environment

_No response_

### Anything else?

_No response_

huanghaoyuanhhy:
close #117

### DIFF
--- a/core/backup_context.go
+++ b/core/backup_context.go
@@ -615,7 +615,10 @@ func (b BackupContext) executeCreateBackup(ctx context.Context, request *backupp
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


yelusion2 (approved):
/lgtm

PR #336, opened 2024-05-10
Update Dockerfile

erigo:
Add MinIO Client for migrating data in each Milvus cluster.


wayblink:
@erigo Hi, Thanks for contribution. Why we need that?

wayblink:
@erigo I get your point, good for now. We will support cross storage backup in our roadmap. Then you will not need to use mc client to copy to local. How about the image size change after you add mc client.

erigo:
image size from 41.7MB to 68.2MB.

wayblink:
@erigo Merged.

PR #346, opened 2024-05-15
test: add sparse vector datatype

zhuwenxing:
/kind improvement


XinyuDu:
Which version will support sparse vector field backup？

PR #347, opened 2024-05-22
Fix enabling both drop_exist_collection and skip_create_collection 

syang1997:
If skip_create_collection is not enabled, the collection will be validated before restoration. Therefore, when using drop_exist_collection, skip_create_collection must be enabled first. However, enabling both skip_create_collection and drop_exist_collection will result in the collection being deleted without being correctly recreated afterwards.

### DIFF
--- a/cmd/restore.go
+++ b/cmd/restore.go
@@ -89,7 +89,7 @@ var restoreBackupCmd = &cobra.Command{
 			MetaOnly:             restoreMetaOnly,
 			RestoreIndex:         restoreRestoreIndex,
 			UseAutoIndex:         restoreUseAutoIndex,
-			DropExistCollection:  restoreDropExistIndex,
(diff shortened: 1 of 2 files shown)


syang1997:
up

syang1997:
If skip_create_collection is not enabled, the collection will be validated before restoration. Therefore, when using drop_exist_collection, skip_create_collection must be enabled first. However, enabling both skip_create_collection and drop_exist_collection will result in the collection being deleted without being correctly recreated afterwards.

syang1997:
If skip_create_collection is not enabled, the collection will be validated before restoration. Therefore, when using drop_exist_collection, skip_create_collection must be enabled first. However, enabling both skip_create_collection and drop_exist_collection will result in the collection being deleted without being correctly recreated afterwards.

PR #348, opened 2024-05-22
Fix enabling both drop_exist_collection and skip_create_collection 

syang1997:
If skip_create_collection is not enabled, the collection will be validated before restoration. Therefore, when using drop_exist_collection, skip_create_collection must be enabled first. However, enabling both skip_create_collection and drop_exist_collection will result in the collection being deleted without being correctly recreated afterwards.

### DIFF
--- a/cmd/restore.go
+++ b/cmd/restore.go
@@ -89,7 +89,7 @@ var restoreBackupCmd = &cobra.Command{
 			MetaOnly:             restoreMetaOnly,
 			RestoreIndex:         restoreRestoreIndex,
 			UseAutoIndex:         restoreUseAutoIndex,
-			DropExistCollection:  restoreDropExistIndex,
+			DropExistCollection:  restoreDropExistCollection,
 			DropExistIndex:       restoreDropExistIndex,
 			SkipCreateCollection: restoreSkipCreateCollection,
 		})
--- a/core/backup_impl_restore_backup.go
+++ b/core/backup_impl_restore_backup.go
@@ -438,16 +438,27 @@ func (b *BackupContext) executeRestoreCollectionTask(ctx context.Context, backup
 	}
 
 	if task.GetDropExistCollection() {
-		err := b.milvusClient.DropCollection(ctx, targetDBName, targetCollectionName)
+		//check if the collection exist, if collection exist, will drop it
+		exist, err := b.milvusClient.HasCollection(ctx, targetDBName, targetCollectionName)
 		if err != nil {
-			errorMsg := fmt.Sprintf("fail to drop collection, CollectionName: %s.%s err: %s", targetDBName, targetCollectionName, err)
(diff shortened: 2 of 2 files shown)


syang1997:
> @syang1997 Hi, thanks for contribute. skip_create_collection and drop_exist_collection is designed not to use at the same time. In your pr, drop_exist_collection has a higher priority than skip_create_collection. This avoid the unexpected situation you mentioned. I prefer to add a parameter check, refuse the request if user set skip_create_collection and drop_exist_collection both true.

However, the current situation is that drop_exist_collection is not taking effect. In the scenario where a collection already exists in the cluster, setting drop_exist_collection to true and skip_create_collection to false, the restoration process fails due to the code snippet at (https://github.com/zilliztech/milvus-backup/blob/v0.4.13/core/backup_impl_restore_backup.go#L263) not passing the check. Therefore, in practice, drop_exist_collection has no actual effect.

syang1997:
> @syang1997 Hi, thanks for contribute. skip_create_collection and drop_exist_collection is designed not to use at the same time. In your pr, drop_exist_collection has a higher priority than skip_create_collection. This avoid the unexpected situation you mentioned. I prefer to add a parameter check, refuse the request if user set skip_create_collection and drop_exist_collection both true.



> @syang1997 Hi, thanks for contribute. skip_create_collection and drop_exist_collection is designed not to use at the same time. In your pr, drop_exist_collection has a higher priority than skip_create_collection. This avoid the unexpected situation you mentioned. I prefer to add a parameter check, refuse the request if user set skip_create_collection and drop_exist_collection both true.

Therefore, I modified the code in this way to make both drop_exist_collection and skip_create_collection take effect simultaneously. This ensures that the collection is deleted while also preventing the issue of the collection not being recreated after deletion due to skip_create_collection.

syang1997:
> @syang1997 Hi, thanks for contribute. skip_create_collection and drop_exist_collection is designed not to use at the same time. In your pr, drop_exist_collection has a higher priority than skip_create_collection. This avoid the unexpected situation you mentioned. I prefer to add a parameter check, refuse the request if user set skip_create_collection and drop_exist_collection both true.

@wayblink If it is required to enable drop_exist_collection and skip_create_collection independently, then the condition where drop_exist_collection is true should be assigned the capability of skip_create_collection (collection existence not being an error, and collection creation taking place after the deletion process). In this case, it becomes similar to setting skip_create_collection to true as well.

syang1997:
In the code flow, there is a check controlled by skip_create_collection before the deletion operation of drop_exist_collection, and a collection creation controlled by skip_create_collection after the deletion. I understand that the correct functionality of drop_exist_collection should be: if a collection exists in the cluster before the restoration, it will first be deleted, and then a new one will be created before proceeding with the restoration process. With the current design, it is impossible to achieve the envisioned scenario where drop_exist_collection and skip_create_collection are not enabled simultaneously.

wayblink:
@syang1997 OK, make sense.

PR #418, opened 2024-09-03
BugFix: fix backup path parse issue

balloon1995:


balloon1995:
Cannot execute bulk insert task correctly. There's a repeated number string in the path:
image
image

wayblink:
@balloon1995 Thanks! I prefer to address this in the upper logic rather than change the path in `copy` method. fixed in #422

Issue #484, opened 2024-12-20
[Bug]: For collections with full text search enabled, after backup and restoration, the function attributes are missing in the schema


zhuwenxing:
### Current Behavior

![image](https://github.com/user-attachments/assets/1c0498e9-45d9-4f6e-8381-1bdad9d47eac)

### Anything else?

failed ci job:https://github.com/zilliztech/milvus-backup/actions/runs/12424891542/job/34690922061?pr=447

huanghaoyuanhhy:
All the test cases added in #447 have passed

PR #492, opened 2025-01-06
feat: support Milvus 2.5 function feature by bypassing SDK  

huanghaoyuanhhy:
Milvus 2.5 introduced the function feature, which is not well-supported by the older SDK versions. The new SDK hides many internal IDs used by the backup tool. To address this, the backup tool now interacts with Milvus directly via gRPC, bypassing the SDK.  

related to: #484


Thomas-HuWei (approved):
/lgtm

PR #495, opened 2025-01-06
GCS native support (GCS implemented using Google Cloud Storage libraries)

gifi-siby:
Native support for Google cloud storage using the Google Cloud Storage libraries. Authentication is performed using GCS service account credentials JSON.

Milvus supports GCS native and Milvus-Backup doesn't. Since Milvus supports GCS native, this enhancement is needed.

Related Issue: [#444 ](https://github.com/zilliztech/milvus-backup/issues/444)

PRs worked together are:
[#494 ](https://github.com/zilliztech/milvus-backup/pull/494)
[#496 ](https://github.com/zilliztech/milvus-backup/pull/496)

PR #494, opened 2025-01-06
TLS Authentication in Backup Tool

gifi-siby:
Milvus-Backup currently doesn't support TLS and thus, I am not able to connect using certificates.
I would like to add TLS support for Milvus-Backup tool

Related issue: [#493 ](https://github.com/zilliztech/milvus-backup/issues/493)

PRs worked together are:
[#495 ](https://github.com/zilliztech/milvus-backup/pull/495)
[#496 ](https://github.com/zilliztech/milvus-backup/pull/496)

Issue #446, opened 2024-10-26
[Bug]: milvus-backup failed for Azure Blob Storage Account


gifi-siby:
### Current Behavior

When I try to backup using azure, I am getting the below error. 
Using same azure bucket for milvus storage and backup storage. 
`["CopyFromURL err"] [error="PUT https://<container-name>.blob.core.windows.net:443/<bucket-name>/devfilesBackup/Backk/binlogs/insert_log/453399115307976339/453399115307976340/453399115308176349/453399115308176349/0\n--------------------------------------------------------------------------------\nRESPONSE 401: 401 Server failed to authenticate the request. Please refer to the information in the www-authenticate header.\nERROR CODE: CannotVerifyCopySource\n--------------------------------------------------------------------------------\n<?xml version="1.0" encoding="utf-8"?><Error><Code>CannotVerifyCopySource</Code><Message>Server failed to authenticate the request. Please refer to the information in the www-authenticate header.\nRequestId:89115e49-f01e-0058-1310-259a52000000\nTime:2024-10-23T05:54:21.2939119Z</Message></Error>\n--------------------------------------------------------------------------------\n"]`

### Steps To Reproduce

```markdown
# Configures the system log output.
log:
  level: debug # Only supports debug, info, warn, error, panic, or fatal. Default 'info'.
  console: true # whether print log to console
  file:
    rootPath: "logs/backup.log"

http:
  simpleResponse: true

# milvus proxy address, compatible to milvus.yaml
milvus:
  address: localhost
  port: 19530 #443
  authorizationEnabled: true
  tlsMode: 1
  user: "ibmlhadmin"
  password: "password"

# Related configuration of minio, which is responsible for data persistence for Milvus.
minio:
  # Milvus storage configs, make them the same with milvus config
  storageType: "azure"
  # storageType: azure # support storage type: local, minio, s3, aws, gcp, ali(aliyun), azure, tc(tencent), gcpnative
  address: core.windows.net # Address of MinIO/S3
  port: 443 # Port of MinIO/S3
  accessKeyID: my_access_key # accessKeyID of MinIO/S3
  secretAccessKey:my_secret_key # MinIO/S3 encryption string
  useSSL: false # Access to MinIO/S3 with SSL
  useIAM: false
  iamEndpoint: ""
  bucketName: bucket-name # Milvus Bucket name in MinIO/S3, make it the same as your milvus instance
[...]

PR #496, opened 2025-01-06
milvus-backup failure for Azure Blob Storage Account

gifi-siby:
When I try to backup using azure, I am getting multiple errors.
Related issue: [#446 ](https://github.com/zilliztech/milvus-backup/issues/446)

PRs worked together are:
[#494 ](https://github.com/zilliztech/milvus-backup/pull/494)
[#495 ](https://github.com/zilliztech/milvus-backup/pull/495)

### DIFF
--- a/core/storage/azure_chunk_manager.go
+++ b/core/storage/azure_chunk_manager.go
@@ -277,29 +277,54 @@ func (mcm *AzureChunkManager) RemoveWithPrefix(ctx context.Context, bucketName s
 	if err != nil {
 		return err
 	}
-	removeKeys := make([]string, 0)
+	// Group objects by their depth (number of / in the key)
+	groupedByLevel := make(map[int][]string)
+	var maxLevel int
 	for key := range objects {
-		removeKeys = append(removeKeys, key)
+		level := strings.Count(key, "/")
+		groupedByLevel[level] = append(groupedByLevel[level], key)
+		if level > maxLevel {
+			maxLevel = level
+		}
 	}
-	i := 0
-	maxGoroutine := 10
-	for i < len(removeKeys) {
-		runningGroup, groupCtx := errgroup.WithContext(ctx)
-		for j := 0; j < maxGoroutine && i < len(removeKeys); j++ {
-			key := removeKeys[i]
-			runningGroup.Go(func() error {
-				err := mcm.removeObject(groupCtx, bucketName, key)
-				if err != nil {
-					log.Warn("failed to remove object", zap.String("bucket", bucketName), zap.String("path", key), zap.Error(err))
-					return err
-				}
-				return nil
-			})
-			i++
+	for level := maxLevel; level >= 0; level-- {
+		// Get the objects at this level
+		keysAtLevel, exists := groupedByLevel[level]
+		if !exists || len(keysAtLevel) == 0 {
+			continue
 		}
-		if err := runningGroup.Wait(); err != nil {
-			return err
+
+		// Dynamically adjust maxGoroutines based on the number of objects at this level
+		maxGoroutines := 10
+		if len(keysAtLevel) < maxGoroutines {
+			maxGoroutines = len(keysAtLevel)
 		}
+		i := 0
+		for i < len(keysAtLevel) {
+			runningGroup, groupCtx := errgroup.WithContext(context.Background())
+			for j := 0; j < maxGoroutines && i < len(keysAtLevel); j++ {
+				key := keysAtLevel[i]
+				runningGroup.Go(func(key string) func() error {
+					return func() error {
+						err := mcm.removeObject(groupCtx, bucketName, key)
+						if err != nil {
+							log.Warn("failed to remove object", zap.String("bucket", bucketName), zap.String("path", key), zap.Error(err))
+							return err
+						}
+						return nil
+					}
+				}(key))
+				i++
+			}
+			if err := runningGroup.Wait(); err != nil {
+				return err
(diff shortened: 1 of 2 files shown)


huanghaoyuanhhy:
It seems that a pager created using `NewListBlobsFlatPager` should not perceive any hierarchical structure, as it is designed to return a flat list of blobs.  I’m not entirely sure why this error is occurring, so could you provide more details？

gifi-siby:
The encountered issues are described in the issue raised [#446 ](https://github.com/zilliztech/milvus-backup/issues/446). I have also included the configurations I used. Could you please review them and let me know if any additional details are needed?

huanghaoyuanhhy:
We have a similar implementation for deletion on Azure, but I haven’t encountered any issues like this. Moreover, Milvus doesn’t seem to use hierarchical deletion either. This error might be caused by other reasons. Let me double-check and get back to you.

huanghaoyuanhhy:
Still trying to reproduce the problem in the original issue.

huanghaoyuanhhy:
This issue is expected to be resolved in the latest version released at the end of March.

huanghaoyuanhhy:
Thank you for your contribution, and I sincerely apologize for the delay in reviewing this PR. 

Regarding the `DirectoryIsNotEmpty` error, I conducted some tests and found that when the hierarchical namespace is enabled, ListBlobsFlatPager includes directories. Based on [discussions](https://github.com/Azure/azure-sdk-for-go/issues/18921#issuecomment-1343903887) in the Azure SDK, Microsoft does not guarantee the compatibility of the `azblob` SDK with storage accounts that have hierarchical namespaces enabled. Given this limitation, a better approach might be to introduce a new `storageType` in Milvus-Backup that utilizes the [azdatalake](https://github.com/Azure/azure-sdk-for-go/tree/main/sdk/storage/azdatalake) SDK at the lower level.

For the copy operation error, I agree that the modification in the PR is necessary—passing the SAS token under all conditions is indeed the correct approach.

That said, while the changes to the copy function should be retained, the logic for hierarchical deletion and file type checking may not be necessary. A more effective solution would be either introducing a new storage type or ensuring that hierarchical namespace remains disabled in the Azure storage account.

gifi-siby:
Thank you for your detailed insights.
Regarding the two options, could you kindly clarify which one would be preferable moving forward? Should I proceed with making the changes, or do you have plans to implement them on your side?
I’m happy to assist with either approach, depending on what works best.
Looking forward to your guidance.

huanghaoyuanhhy:
I think disabling the hierarchical namespace in Azure would be a better approach. Milvus itself does not rely on the hierarchical namespace to function, so keeping it consistent with Milvus would be the preferable choice.

Would you happen to have any specific reasons for enabling the hierarchical namespace on your side?

huanghaoyuanhhy:
Would it be possible to keep only the changes to the copy method in this PR and remove the file type checking part?
Once that’s done, we can proceed with merging it. 

Thanks again for your contribution and patience!

gifi-siby:
@huanghaoyuanhhy  Can you please review the changes

Rachit-Chaudhary11:
@huanghaoyuanhhy we are facing the issue while taking a backup from minio to azure blob storage. As per the earlier comment this fix was supposed to go in on 31st March or so, do you have any ETA?

huanghaoyuanhhy:
@gifi-siby Just wanted to check if you have any plans to update the PR soon? That would help us plan the next steps.

gifi-siby:
@huanghaoyuanhhy I've made the suggested changes and should I update the branch to the latest? Is that expected? If so, I can make those changes soon

gifi-siby:
@huanghaoyuanhhy Can you please review the PR and help me merge it?

huanghaoyuanhhy:
Is there a specific reason for using NewSharedKeyCredential here? In some environments, especially when using Kubernetes ServiceAccount with Workload Identity, the AK/SK might be empty and authentication is done via WorkloadIdentityCredential.

In such cases, GetUserDelegationCredential is more appropriate, as it supports this authentication method.

huanghaoyuanhhy:
Also, please add a [DCO](https://developercertificate.org/) sign-off to your commit(s) so that the bot can remove the `need-dco` label. The sign-off should follow this format:

```
This is my commit message

Signed-off-by: Random J Developer <random@developer.example.org>
```

gifi-siby:
> As I mentioned earlier, in some setups (like when using Kubernetes ServiceAccount with `WorkloadIdentity`), AK/SK might be empty, and `WorkloadIdentityCredential` is used instead. In these cases, `GetUserDelegationCredential` is preferred since it supports that auth method.
> 
> Just wondering — is there a specific reason for using `NewSharedKeyCredential` in this PR?

Thanks for pointing that out! There’s no specific reason for using NewSharedKeyCredential in this PR — I used it mainly because it worked for my setup. But you're absolutely right, GetUserDelegationCredential would be more appropriate in environments using WorkloadIdentity. Happy to adjust that!

gifi-siby:
@huanghaoyuanhhy I have made the modifications and could you please review

huanghaoyuanhhy:
@Rachit-Chaudhary11 
The new version has been released — thanks to @gifi-siby for the contribution! Apologies that the fix didn’t make it into the previous release.

EmilKarimovEPam:
Hello everyone, 

Could you please tell me how to fix the issue. I am struggling on this issue for couple days. WOuld love for any support. Thanks.

PR #507, opened 2025-01-16
test: add index check after restore

zhuwenxing:
/kind improvement


huanghaoyuanhhy:
/lgtm 
Due to the job name change, the ci-passed tag will not be automatically added. This will be fixed in a future commit.

PR #575, opened 2025-03-19
core, client: improve log readability during the flush process

### LINKED ISSUE #537: [Enhancement]: Flush fails and retries, but logs at error level, which may cause confusion

Adjust the log level to warn or info when flush fails but will be retried. Only log at error level if all retries have failed.

### Why is this needed?

Currently, when flush fails, the system retries the operation, but the log level is set to error. This may mislead users into thinking that the flush operation has completely failed, rather than the system automatically retrying, making issue troubleshooting more difficult.

### Anything else?

_No response_

### LINKED ISSUE #548: [Bug]: Backup Creation Lacks Feedback on Flush Process
### Current Behavior

Backups are taking an excessively long time to create and appear to be stalled. 
Investigation of the Milvus access logs reveals the process has been stuck in a flush state for nearly an hour. 
However, the backup log does not show any issues related to the flush operation.

### Expected Behavior

We need the backup tool to provide better feedback on the flush process. Specifically, it should either show the flush is in progress or alert us with a warning if the flush takes longer than an hour.

### Steps To Reproduce

```markdown

```

### Environment

```markdown

```

### Anything else?

_No response_

huanghaoyuanhhy:
fix #548 , fix #537

### DIFF
--- a/core/backup_impl_create_backup.go
+++ b/core/backup_impl_create_backup.go
@@ -398,7 +398,7 @@ func (b *BackupContext) backupCollectionPrepare(ctx context.Context, backupInfo 
 			zap.Int("segmentNumBeforeFlush", len(segmentEntitiesBeforeFlush)))
 		flushResp, err := b.getMilvusClient().Flush(ctx, collectionBackup.GetDbName(), collectionBackup.GetCollectionName())
 		if err != nil {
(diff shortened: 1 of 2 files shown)


Thomas-HuWei (approved):
/lgtm

Issue #476, opened 2024-12-12
[Bug]: Successfully created the backup but failing in restore mid way ( v2.4.4 )


tarunpantqb:
### Current Behavior

I've taken the backup from milvus running in AWS account S3 bucket and copied the backup data into another s3 bucket of different aws account. Now, while running the restore process for second milvus instance, I get below error : 

```
main [2024/12/12 10:37:56.456 +00:00] [INFO] [core/backup_impl_restore_backup.go:850] [getBackupPartitionPaths] [bucketName=lsai-dev-milvus-storage] [backupPath=backup/backup_2024_12_11_09_35_07_561872490] [p │
│ main [2024/12/12 10:37:56.467 +00:00] [ERROR] [core/backup_impl_restore_backup.go:357] ["executeRestoreCollectionTask failed"] [TargetDBName=default] [TargetCollectionName=publications_test] [error="bulk inse │
│ main [2024/12/12 10:37:56.467 +00:00] [ERROR] [core/backup_impl_restore_backup.go:321] ["execute restore collection fail"] [backupId=356c4cfd-b7a3-11ef-95c7-fe22564fb5a1] [error="workerpool: execute job bulk  │
│ main workerpool: execute job bulk insert fail, info: misaligned binlog count, field1:87, field0:85: importing data failed                                                                                        │
│ main duration:3328 s                                                                                                                                                                                             │
│ main time="2024-12-12T10:37:56.572Z" level=info msg="sub-process exited" argo=true error="<nil>" '

```

This comes midway of restore job run, it fails after restoring certain records.
I'm using V2.4.4 for both source and target milvus instances.

huanghaoyuanhhy:
Thank you for your report! 
It seems that some files may be missing during the restore process, which could be related to how the data was copied.

Could you please clarify how you performed the copy operation? It would be helpful to know the method you used. In the meantime, we recommend verifying the number of files in both the source and destination S3 buckets to ensure they match. You could also try the copying process again to ensure all files are correctly transferred.

If the issue persists, please let us know, and we'll be happy to assist you further.

tarunpantqb:
Thanks @huanghaoyuanhhy for your response. 
Both milvus are running in 2 different AWS accounts.

- I ran milvus backup in source AWS account.
- Copied the backup bucket data from source to destination AWS account
- Ran milvus restore in destination AWS account using the bucket that has copied data.

I used aws sync for copying the data.I verified the number of objects and size of buckets. Its same.

[...]

Issue #251, opened 2023-12-06
[Bug]:  When I restored data, some DBs where missing


anikmondavis:
### Current Behavior

When I restored data some DB where missing, one thing is that these DB where empty. But when we restore it should be even an empty db should restored

### Expected Behavior

Should restore all DBs

### Steps To Reproduce

```markdown
./milvus-backup restore -n milvus_backup_dev --restore_index true _recover
```

### Anything else?

is there any method, can specify which folder should restore?

wayblink:
@anikmondavis Hi, thanks. It is a nice catch. Backup metas are designed to be based on collections because we didn't support database at that time. If a database is empty, it will not be backuped and restored. It is a good suggestion to backup empty database as well. We will work on it.

PR #609, opened 2025-04-03
*: read binlog file list from proxy node

huanghaoyuanhhy:
fix: #476 , fix: #251


zhuwenxing:
I haven't looked at the code specifically, just read the title, and I think reading binlog from proxy seems a bit heavy-handed. One of the core principles of Milvus backup is that it shouldn't affect normal read and write traffic. If backup and restore operations consume proxy bandwidth, it would create such an impact.

huanghaoyuanhhy:
>  I think reading binlog from proxy seems a bit heavy-handed. 

You’re right that the title might have been a bit misleading—I’ve updated it. This PR doesn’t actually change the way binlog data is read; it only ensures that the correct binlog file list is obtained from the proxy node. Previously, all binlog files were copied indiscriminately, but not all binlog files in object storage are necessarily valid. Copying invalid files could lead to restore failures.

Since this PR involves a fundamental change to the backup process, the modifications might be relatively large in scope. You can refer to [this issue](https://github.com/milvus-io/milvus/issues/40341) for more context.

zhuwenxing:
OK, I got it.

It was binlog files name list, rather than the files objects. am i right?

huanghaoyuanhhy:
> It was binlog files name list, rather than the files objects.

Yes.

lentitude2tk (approved):
/lgtm

Issue #444, opened 2024-10-21
[Enhancement]: GCS native support (GCS implementation using Google Cloud Storage libraries)


gifi-siby:
### What would you like to be added?

As per [#36212](https://github.com/milvus-io/milvus/issues/36212), milvus is has native support for Google cloud storage using the Google Cloud Storage libraries (Authentication is performed using GCS service account credentials JSON). 
So, MilvusBackup also should support GCS native storage for backup and restore.

### Why is this needed?

Milvus currently supporting GCS native also

xiaofan-luan:
Thanks or the issue, any pr for this?

gifi-siby:
No. I am working on this. Created this ticket to check if there is plan for GCS support

gifi-siby:
Any update on this issue? [#446 ](https://github.com/zilliztech/milvus-backup/issues/446)

PR #564, opened 2025-03-05
gcpnative support for backup tool (GCS implemented using Google Cloud Storage libraries)

gifi-siby:
Native support for Google cloud storage using the Google Cloud Storage libraries. Authentication is performed using GCS service account credentials JSON.

Milvus supports GCS native and Milvus-Backup doesn't. Since Milvus supports GCS native, this enhancement is needed.

Related Issue: https://github.com/zilliztech/milvus-backup/issues/444

Older PR https://github.com/zilliztech/milvus-backup/pull/495 closed

PR #625, opened 2025-04-21
gcpnative support for backup tool (GCS implemented using Google Cloud Storage libraries)

gifi-siby:
Native support for Google cloud storage using the Google Cloud Storage libraries. Authentication is performed using GCS service account credentials JSON.

Milvus supports GCS native and Milvus-Backup doesn't. Since Milvus supports GCS native, this enhancement is needed.

Related Issue: https://github.com/zilliztech/milvus-backup/issues/444


gifi-siby:
@huanghaoyuanhhy Regarding the last [comment](https://github.com/zilliztech/milvus-backup/pull/564#issuecomment-2817722860),
I used `io.Copy` here because we might not always be copying within the same bucket, and `CopierFrom` in GCS has some limitations in cross-bucket or cross-project copies depending on permissions and configurations.

Additionally, this approach gives more control over the data stream, logging, and potential transformations if needed later. If we’re confident about the buckets being compatible, we can definitely switch to CopierFrom for more efficiency. Open to adjusting it based on what's preferred here.

huanghaoyuanhhy:
Thanks for the explanation!

The cross-bucket copy scenario is actually already handled within our storage.Copier implementation. It uses a similar approach to what you did with io.Copy in your PR, but it's encapsulated to maintain consistency.

According to the interface semantics of storage.ChunkManager, the Copy method is expected to rely on the object storage's native copy API rather than performing a manual read/write. This helps keep the implementation aligned with the intended abstraction and allows for backend-specific optimizations.

PR #658, opened 2025-05-20
cmd: export v2 restore flag

### LINKED ISSUE #655: Support multi-Segment merged restore via command-line
Currently, in Milvus backup restore, each Segment triggers a separate BulkInsert job. When restoring a Collection with a large number of Segments, this results in the creation of a large number of jobs. The overhead of job creation and scheduling becomes significant, often outweighing the actual data processing time.

We've already implemented support for restoring multiple Segments in a single BulkInsert job (v2 restore) in the backup. However, this feature is currently not available via the command-line interface and is not yet documented.

In internal testing, we've observed up to 10x performance improvements when using this mode to restore large Collections.

TODO:

- Add a --use_v2_restore flag to enable this feature from the CLI
- In the future, make it the default behavior
- Provide --use_v1_restore to allow fallback to the original per-Segment job mode
- Document both restore modes and their trade-offs

huanghaoyuanhhy:
fix #655


yelusion2 (approved):
/lgtm

Issue #695, opened 2025-06-12
[Proposal]: Introduce --filter to Replace Multiple Complex Parameters


huanghaoyuanhhy:
# Description
Currently, when users need to backup or restore specific collections, users can use three parameters: `--collections`, `--database_collections`, and `--databases`. This design presents the following issues:

- The priority of parameters is unclear, leading to confusion.
- The usage method is complex, increasing the learning curve for users.
- The logical relationship between parameters is not clear, which can lead to errors.

# Proposal

Introduce a unified way to specify which collections should be backup or restore. The specific design is as follows:

- Use a dot `.` to separate the database and Collection, formatted as `db.coll`.
- Use commas `,` to separate multiple `db.coll` combinations, for example, `db1.coll1,db2.coll2`.
- Use `*` as a wildcard, but the first phase will only support complete Collection wildcards, such as `db1.*`, and will **NOT** support `db*.coll1` or `db.coll*`.
- Use `db.` to indicate that only the specified database should be restored, without restoring any collections under that database.

# Transition Period
To ensure a smooth transition, the old parameters `--collections`, `--database_collections`, and `--databases` will be retained for a period. If the user passes the `--filter` parameter, the system will prioritize using `--filter` and ignore the old parameters. In the future, the old three parameters will be removed.

PR #726, opened 2025-07-04
*: abstract filter and mapping logic into planner for task generation

huanghaoyuanhhy:
This refactors the restore process by extracting the filtering and mapping steps into a dedicated planner component. The planner generates restore tasks based on the restore plan, making the pipeline more modular.

relate: #695


lentitude2tk (approved):
/lgtm

Issue #629, opened 2025-04-24
Faster Backup and Restore


huanghaoyuanhhy:
There are several performance bottlenecks in the current Milvus backup and restore workflow, especially noticeable in scenarios involving a large number of small segments. This issue documents the known problems and potential directions for optimization.

Known issues (by priority):

* **~~Restore with crossStorage flag is not concurrent~~**
When the `crossStorage` flag is enabled, data copying is performed sequentially. Multithreaded copying should be used to speed up the process.
Fixed by multiple pr

* **~~Each segment is restored as an individual job~~**
Restoring many small segments results in overhead due to the large number of restore jobs. These can be grouped into fewer jobs to improve efficiency.
fixed: [Multi-Segment Restore Guide](https://github.com/zilliztech/milvus-backup/blob/main/docs/user_guide/mul_seg_restore.md
)

* **~~Segment-level file copy during backup is sequential~~**
During backup, segments are copied one by one in sequence. Although each segment's internal files (e.g., binlogs, deltas, stats) are copied in parallel, the overall backup process is bottlenecked by the sequential handling of segments. Enabling concurrent copying across segments can significantly improve performance, especially when dealing with a large number of small segments.
Fixed: #729 

* **~~Restore concurrency is based on collection-level parallelism and partition-level serialization~~**
Task parallelism is currently coarse-grained: collections are restored in parallel, but partitions are serialized. This can lead to underutilized compute resources. Concurrency should be determined based on binlog size or segment granularity for better scheduling.
Fixed #733 

* **~~Flush during backup is done collection by collection~~**
Backup currently triggers flushes one collection at a time. This causes unnecessary delays when multiple collections are involved. A bulk or global flush mechanism is needed.
Fix: #856 #852

huanghaoyuanhhy:
milvus improvement https://github.com/milvus-io/milvus/pull/41552

sonakshireddym:
> Restore concurrency is based on collection-level parallelism and partition-level serialization
> Task parallelism is currently coarse-grained: collections are restored in parallel, but partitions are serialized. This can lead to underutilized compute resources. Concurrency should be determined based on binlog size or segment granularity for better scheduling.

I'm testing out this tool to restore one collection consisting of around 160 million docs and its taking around 16 hours with the current code (v0.5.5). Its taking <2 hours if the partitions are run in parallel. Is there any plan to make changes to partition concurrency in the next version? 
[...]

PR #729, opened 2025-07-07
backup: support parallel backup of segments

huanghaoyuanhhy:
Previously, only collection-level backup was parallelized, which helped in environments with many small collections. However, in cases with few but large collections, backup speed was limited by sequential segment processing. This change enables segment-level parallelism to improve throughput in such scenarios.

relate: #629


lentitude2tk (approved):
/lgtm

PR #739, opened 2025-07-17
cloud, migrate: use new migrate api

huanghaoyuanhhy:


huanghaoyuanhhy:
some milvus bug blcok the ci

PR #762, opened 2025-08-29
storage: use original context for semaphore acquisition

huanghaoyuanhhy:
Prevent semaphore errors from masking actual copy failures by using the original context instead of errgroup's cancelable context for acquisition. Maintains proper cancellation for operations while ensuring semaphore waits aren't interrupted by sibling failures.


Thomas-HuWei (approved):
/lgtm

Issue #742, opened 2025-07-21
[Bug]: backup state_code bug


rere950303:
### Current Behavior

- milvus version: v2.4.23
- backup version: v.0.5.7

--- 
 
- This issue is related to https://github.com/zilliztech/milvus-backup/issues/735.
- As per your guidance, we are testing the latest version of the backup tool, v0.5.7. When performing the initial backup, it completes successfully as shown in the attached screenshot, and we can confirm that the backup files, including metadata, are properly stored in the object storage.

<img width="435" height="145" alt="Image" src="https://github.com/user-attachments/assets/8afb0297-818e-4262-8b17-b6ec49425dd4" />

- However, when the backup pod is restarted, the backup `state_code` changes to 1 (`BACKUP_EXECUTING`), as shown below. There are no changes in the object storage.

<img width="408" height="129" alt="Image" src="https://github.com/user-attachments/assets/763d7172-4fd5-410d-b2d7-c657026bd51a" />

### Expected Behavior

Even if the Kubernetes backup pod is restarted, the `state_code` of previously completed backups should remain as 2 (`BACKUP_SUCCESS`).
This bug did not exist in backup tool version v0.5.2.

### Steps To Reproduce

```markdown

```

### Environment

```markdown

```

rere950303:
@huanghaoyuanhhy hello! could you check this issue plz? 🙏

rere950303:
Looking at the `full_meta.json`, the state_code is set to 1 even though the backup was completed.

huanghaoyuanhhy:
Thanks for the detailed feedback.

OK, I will work on fixing this issue promptly. However, please note that the status_code field in` full_meta.json` should not be relied upon — I plan to deprecate this field in a future release.

rere950303:
@huanghaoyuanhhy 
- Thank you for your response. In that case, which field should we check to determine whether the backup has been completed?

---

- Just to briefly explain our backup tool usage scenario for your reference:
- We perform backups daily at midnight and retain the three most recent backup copies. However, if the `state_code` of the most recent backup is 1 (i.e., still in progress), our logic is designed to log a warning and skip the backup process.

huanghaoyuanhhy:
The status in the HTTP API response is reliable and should be used to determine the backup state. I'm still investigating the root cause of the issue you encountered. Will keep you updated once I have more findings.

rere950303:
@huanghaoyuanhhy 
Could you please let me know the current status?

[...]

PR #784, opened 2025-09-13
fix: ensure final backup state is always persisted

rere950303:
fix https://github.com/zilliztech/milvus-backup/issues/742

### DIFF
--- a/cmd/restore/restore_test.go
+++ b/cmd/restore/restore_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
 	"github.com/zilliztech/milvus-backup/core/restore"
 )
 
--- a/core/backup_impl_create_backup.go
+++ b/core/backup_impl_create_backup.go
@@ -113,43 +113,74 @@ func (b *BackupContext) CreateBackup(ctx context.Context, request *backuppb.Crea
 	}
 }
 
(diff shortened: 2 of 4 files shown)


rere950303:
@huanghaoyuanhhy 
I would greatly appreciate it if you could take a look whenever you have time 🙏

huanghaoyuanhhy:
Thanks for the PR! The fix addresses the symptom, but the root issue is that Backup task state and Backup metadata are coupled.

The *_meta.json is only written after a successful backup, so its status/progress is always success and 100%, meaning task state doesn’t belong there. The current HTTP get backup API also exposes too much detail, leading to either coarse-grained locking or unsafe concurrent access.

We had a similar issue in restore, which I’ve already solved by moving task progress management into the taskmgr package. I’d like to take the same approach for backup, but haven’t started yet.

rere950303:
> Thanks for the PR! The fix addresses the symptom, but the root issue is that Backup task state and Backup metadata are coupled.
> 
> The *_meta.json is only written after a successful backup, so its status/progress is always success and 100%, meaning task state doesn’t belong there. The current HTTP get backup API also exposes too much detail, leading to either coarse-grained locking or unsafe concurrent access.
> 
> We had a similar issue in restore, which I’ve already solved by moving task progress management into the taskmgr package. I’d like to take the same approach for backup, but haven’t started yet.

Thank you for the detailed explanation! I also thought that the presence of the meta file meant the backup was successful. As you mentioned, since you’ll be providing a fix soon, I’ll go ahead and close this PR.
Thanks again 🙇

PR #785, opened 2025-09-15
storage, proto: format code

huanghaoyuanhhy:


huanghaoyuanhhy:
There are still some problems with obtaining the Azure key, which can be ignored for now.

PR #788, opened 2025-09-18
configs, log: use `filename` instead of `rootPath`

huanghaoyuanhhy:
configs: split rootPath and filename


Thomas-HuWei (commented):
/lgtm

PR #832, opened 2025-11-10
storage: use multipart copy when object size too big

### LINKED ISSUE #831: [Enhancement]: Support Multipart Copy for Large Object Transfers

Introduce Multipart Copy support to improve large file handling performance and stability.

ref S3 [doc](https://docs.aws.amazon.com/AmazonS3/latest/API/API_UploadPartCopy.html)

### Why is this needed?

After Milvus v2.6 introduced Storage v2, the size of binlog files has significantly increased due to the new data layout and compression behavior.
As a result, Milvus-backup may encounter performance bottlenecks or transfer failures when copying large files using the current single-request `CopyObject` approach.

The existing implementation uses a single `CopyObject` operation for file replication (e.g., during backup migration or restore).
This approach causes several issues when handling large files:

- Slow transfer speed due to lack of parallelism.
- Request timeouts or connection interruptions during long copy operations.
- Incompatibility with some S3-compatible storages that limit single-copy object size (e.g., 5 GB).

### Anything else?

_No response_

huanghaoyuanhhy:
fix #831

### DIFF
--- a/internal/storage/aliyun.go
+++ b/internal/storage/aliyun.go
@@ -27,12 +27,7 @@ func newAliyunClient(cfg Config) (*MinioClient, error) {
 		return nil, fmt.Errorf("storage: aliyun unsupported credential type: %s", cfg.Credential.Type.String())
 	}
 
-	cli, err := minio.New(cfg.Endpoint, &opts)
-	if err != nil {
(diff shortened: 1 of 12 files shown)


lentitude2tk (approved):
/lgtm

Issue #767, opened 2025-09-01
[Enhancement]: Support batch flush before backup


huanghaoyuanhhy:
### What would you like to be added?

Milvus Backup tool needs to flush collections before starting a backup.  
Currently:
- It must call flush for each collection separately.  
- Flush operations are rate-limited, so this becomes very slow when there are many collections.  

To solve this, we need the Milvus server to support **flushing multiple collections (across databases) in one request**.

### Why is this needed?

- Reduce RPC overhead.  
- **Greatly improve backup speed in scenarios with a very large number of collections.**

### Anything else?

This issue depends on the feature request in Milvus 
milvus-io/milvus#44156

PR #856, opened 2025-11-27
*: use FlushAll API to speed up flush

huanghaoyuanhhy:
Use Milvus `FlushAll` API to flush all collections in a single call instead of flushing collections one-by-one. This reduces RPC count and lowers total flush latency.

relate: #855 
fix：#767


wenhuiZilliz (approved):
lgtm

lentitude2tk (approved):
/lgtm

PR #894, opened 2025-12-26
restore: Guess dynamic field ID for restore

huanghaoyuanhhy:
This commit introduces a new function `GuessDynFieldID` to intelligently determine the dynamic field ID during the restore process.

The `$meta` field is not explicitly returned by the `describeCollection` API, the dynamic field ID is not readily available. This function addresses this issue by inferring the ID based on the existing field schema.

The guessing logic works as follows:
1.  If a field is explicitly marked with `IsDynamic: true`, its ID is used.
2.  If no field is marked as dynamic, the function searches for gaps in the sequence of field IDs. The first gap found is assumed to be the dynamic field ID. This covers cases where fields have been added.
3.  If there are no gaps in the field IDs, the dynamic field ID is assumed to be the maximum field ID plus one.


lentitude2tk (approved):
/lgtm

PR #901, opened 2026-01-06
[backup]prevent overlapping backup tasks by checking task state

wg1026688210:
prevent overlapping backup tasks by checking task state

fix: #906

### DIFF
--- a/cmd/create/create.go
+++ b/cmd/create/create.go
@@ -9,8 +9,6 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/spf13/cobra"
-	clientv3 "go.etcd.io/etcd/client/v3"
-
 	"github.com/zilliztech/milvus-backup/cmd/root"
 	"github.com/zilliztech/milvus-backup/core/backup"
 	"github.com/zilliztech/milvus-backup/core/paramtable"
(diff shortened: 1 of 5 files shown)


huanghaoyuanhhy:
Thanks for the PR.

I think there are a few points worth reconsidering:
- The task name check is currently done on the server API side, so the cmd interface may still create tasks with the same name.
- Using `GetBackupTaskByName` for a pre-check can lead to read-before-write race conditions under concurrent requests.

It might be better to move the uniqueness check into `taskmgr.(*Mgr).AddBackupTask`, so the validation and creation are handled atomically.

wg1026688210:
> Thanks for the PR.
> 
> I think there are a few points worth reconsidering:
> 
> * The task name check is currently done on the server API side, so the cmd interface may still create tasks with the same name.
> * Using `GetBackupTaskByName` for a pre-check can lead to read-before-write race conditions under concurrent requests.
> 
> It might be better to move the uniqueness check into `taskmgr.(*Mgr).AddBackupTask`, so the validation and creation are handled atomically.

done @huanghaoyuanhhy

wg1026688210:
#906

PR #916, opened 2026-01-22
backup: change cipher detection from cipher.enabled to cipher.ezID/key

huanghaoyuanhhy:
## Summary
- Follow the milvus encryption plugin changes
- Change the cipher detection logic from checking `cipher.enabled` property to checking the presence of both `cipher.ezID` and `cipher.key` properties
- Update related unit tests to cover the new logic

## Test plan
- [x] Unit tests pass (`go test -v ./core/backup/... -run TestDatabaseTask`)

### DIFF
--- a/core/backup/database_task.go
+++ b/core/backup/database_task.go
@@ -12,7 +12,10 @@ import (
 	"github.com/zilliztech/milvus-backup/internal/pbconv"
 )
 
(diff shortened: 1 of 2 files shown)


wenhuiZilliz (approved):
/lgtm

Issue #926, opened 2026-01-29
[Enhancement]: Refactor configuration loading code


huanghaoyuanhhy:
## What would you like to be added?

Refactor the configuration loading code in `core/paramtable/` to improve code quality and fix existing issues.

### 1. Panic Abuse
The current implementation uses `panic()` extensively for error handling:
- `base_table.go:LoadYaml()` - panics on config file errors
- `params.go` - all `Parse*` functions panic on conversion errors
- Configuration parsing failures cause application crashes instead of graceful error handling

### 2. Bugs
- `params.go:402` - Wrong error message: `panic("parse bool useIAM:...")` should be `useSSL`
- `params.go:168-181` - Logic error in `initUser/initPassword`: error handling is overwritten immediately
- `RPCChanelName` typo - should be `RPCChannelName`

### 3. Unused Functions
In `base_table.go`:
- `LoadWithPriority()`, `LoadWithDefault2()`, `LoadRange()`, `ParseFloat()`, `ParseDataSizeWithDefault()` - never called

In `mem_kv.go`:
- `MultiRemoveWithPrefix()` - raises `panic("not implement")`
- `LoadBytes*`, `SaveBytes*`, `MultiLoad*` methods - never used

### 4. Unnecessary MemoryKV
- Uses Google's btree library to implement an in-memory KV store
- Can be entirely replaced by viper which is already used for YAML parsing
- The whole `internal/kv/mem/` package can be removed

### 5. Incorrect Comments
- `CloudProvider` field marked deprecated but still initialized and used

## Why is this needed?

1. **Reliability**: Panic-based error handling causes crashes instead of graceful degradation
2. **Maintainability**: Dead code and unnecessary abstraction make the codebase harder to understand
3. **Correctness**: Existing bugs can cause confusing error messages and unexpected behavior

PR #930, opened 2026-01-30
cfg: replace paramtable with cfg-based config loading

huanghaoyuanhhy:
## Summary
Replace the legacy `core/paramtable` configuration system with the new `internal/cfg` package.

## Changes
- Remove `core/paramtable` and migrate all call sites to `internal/cfg`
- Load config from YAML with flattened keys and typed values (`cfg.Value[T]`)
- Implement override precedence: `--set` (override map) > env > config file > default
- Update docs and add cfg loading tests


huanghaoyuanhhy:
fix #926

PR #934, opened 2026-02-02
backup: add collection listing to Task with filter support

huanghaoyuanhhy:
## Summary
- Optimize collection listing by querying directly based on filter instead of listing all then filtering
- When filter specifies exact collections (db.coll), use HasCollection API to verify existence
- When filter specifies db.*, list collections only for that database
- Remove redundant list-then-filter pattern for better performance

## Changes
- Refactor `listDBAndNSS` to check filter first and choose appropriate strategy
- Add `listAllDBAndNSS` for no-filter case (list all databases and collections)
- Add `listFilteredDBAndNSS` for filtered case (query only needed data)
- Update tests to reflect new behavior


lentitude2tk (approved):
/lgtm

PR #945, opened 2026-02-09
restore: strip backupDir instead of backupRootPath in temp path

huanghaoyuanhhy:
## Summary
- Use `backupDir` instead of `backupRootPath` when stripping prefix in `copyToMilvusBucket`, producing cleaner temp paths
- Remove the now-unused `BackupRootPath` field from `TaskArgs` and `collTaskArgs`

## Changes
- `core/restore/coll_task.go`: replace `backupRootPath` with `backupDir` in `copyToMilvusBucket`, remove `backupRootPath` from both structs
- `core/restore/task.go`: remove `BackupRootPath` from `TaskArgs`, drop assignment in `newCollTask`
- `cmd/restore/restore.go`: remove `BackupRootPath` assignment

/kind improvement

### DIFF
--- a/cmd/restore/restore.go
+++ b/cmd/restore/restore.go
@@ -296,17 +296,16 @@ func (o *options) toArgs(params *cfg.Config) (restore.TaskArgs, error) {
 	}
 
 	return restore.TaskArgs{
-		TaskID:         uuid.NewString(),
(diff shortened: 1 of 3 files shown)


lentitude2tk (approved):
/lgtm

PR #946, opened 2026-02-09
client: prefer connect version in GetVersion for feature gating

huanghaoyuanhhy:
## Summary
In custom/dev Milvus deployments, `GetVersion` RPC may return an empty string when `MILVUS_GIT_BUILD_TAGS` env var is not set. This causes `semver.NewVersion("")` to fail, which makes client initialization fail entirely during `checkFeature()`.

## Changes
- `GetVersion()` now checks `g.serverVersion` (from Connect RPC's `BuildTags`) first — if it's a valid semver, return it directly without issuing the GetVersion RPC
- Falls back to GetVersion RPC only when the connect version is invalid
- Add `TestGrpcClient_GetVersion` (3 cases) tests

/kind improvement

### DIFF
--- a/internal/client/milvus/grpc.go
+++ b/internal/client/milvus/grpc.go
@@ -409,13 +409,20 @@ func (g *GrpcClient) HasFeature(flag FeatureFlag) bool {
 }
 
 func (g *GrpcClient) GetVersion(ctx context.Context) (string, error) {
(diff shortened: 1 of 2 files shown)


lentitude2tk (commented):
/lgtm

Issue #954, opened 2026-02-25
[Bug]: 在获取S3的list列表时报错


hellogitee:
### Milvus-Backup Version

0.5.9-rc1

### Can you reproduce this issue on the latest version?

Yes, it also happens on the latest version

### Current Behavior

使用0.5.9-rc1备份时，报如下错误：
[ERROR] [core/backup_context.go:146] ["failed to initial backup storage client"] [error="create backup storage client: storage: create bucket if not exist storage: s3 list objects Get \"http://xxx.xxx.xxx.xxx/milvus-backup/?delimiter=%2F&encoding-type=url&fetch-owner=true&list-type=2&prefix=\": net/http: timeout awaiting response headers"] [stack="github.com/zilliztech/milvus-backup/core.(*BackupContext).getBackupStorageClient\n\t/home/runner/work/milvus-backup/milvus-backup/core/backup_context.go:146\ngithub.com/zilliztech/milvus-backup/core.(*BackupContext).ListBackups\n\t/home/runner/work/milvus-backup/milvus-backup/core/backup_context.go:281\ngithub.com/zilliztech/milvus-backup/cmd/list.(*options).run\n\t/home/runner/work/milvus-backup/milvus-backup/cmd/list/list.go:26\ngithub.com/zilliztech/milvus-backup/cmd/list.NewCmd.func1\n\t/home/runner/work/milvus-backup/milvus-backup/cmd/list/list.go:46\ngithub.com/spf13/cobra.(*Command).execute\n\t/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1019\ngithub.com/spf13/cobra.(*Command).ExecuteC\n\t/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1148\ngithub.com/spf13/cobra.(*Command).Execute\n\t/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.10.1/command.go:1071\ngithub.com/zilliztech/milvus-backup/cmd.Execute\n\t/home/runner/work/milvus-backup/milvus-backup/cmd/cmd.go:37\nmain.main\n\t/home/runner/work/milvus-backup/milvus-backup/main.go:20\nruntime.main\n\t/opt/hostedtoolcache/go/1.24.7/x64/src/runtime/proc.go:283"]

执行list列表时，报：
[...]

PR #968, opened 2026-03-03
storage: avoid listing all objects in BucketExist

huanghaoyuanhhy:
## Summary
`BucketExist` was iterating through every object in the bucket via `ListObjects` to check bucket existence, which caused unnecessary overhead on large buckets.

## Changes
- Set `MaxKeys=1` in `BucketExist` and return immediately after the first result, making the check O(1) regardless of bucket size
- Add unit tests for `CreateBucketIfNotExist`

Related to #954

### DIFF
--- a/internal/storage/funcs_test.go
+++ b/internal/storage/funcs_test.go
@@ -130,6 +130,33 @@ func TestExist(t *testing.T) {
 	})
 }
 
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/lgtm

PR #987, opened 2026-03-18
restore: warn instead of fail when bulk insert has no progress

huanghaoyuanhhy:
## Summary
- When bulk insert has no progress for 60 minutes, log a warning and keep waiting instead of failing the restore task, consistent with flush behavior.

## Changes
- Remove `return errors.New(...)` on bulk insert timeout in `checkBulkInsertViaGrpc`, `checkBulkInsertViaRestful`, and `checkBulkInsertJob`
- Reset `lastUpdateTime` after warning so the warning repeats every 60 minutes instead of every 3 seconds
- Fix missing `lastProgress` update in progress tracking

/kind improvement


lentitude2tk (approved):
/lgtm

Issue #970, opened 2026-03-05
[Enhancement]: Support customizing Swagger BasePath for reverse proxy


noisrucer:
### What would you like to be added?

Hi, i'd like to propose an enhancement for the Swagger base path.

## Problem
When milvus-backup is deployed behind a reverse proxy with a path prefix (e.g., `/{namespace}/backup/`), the Swagger UI sends requests to the wrong endpoint.

The proxy strips the prefix before forwarding to the backup server, so the server has no knowledge of its external path. Since `SwaggerInfo.BasePath` is hardcoded to `/api/v1`, Swagger UI constructs request URLs as `http://{host}/api/v1/restore` instead of `http://{host}/{namespace}/backup/api/v1/restore`.

## Root Cause

https://github.com/zilliztech/milvus-backup/blob/1acbdfa02b28aeb9233bf1b58c7d789d81818f5b/main.go#L16-L21

`docs.SwaggerInfo.BasePath` is set to `/api/v1` at compile time and there is no way to override it at runtime without modifying the source code.

## Proposed Solution

**Read an environment variable** to allow overriding `SwaggerInfo.BasePath` at startup. In `core/server/server.go`, before the swagger route is registered:

```go
// registerHTTPServer register the http server, panic when failed
func (s *Server) initEngine() {
	if !s.params.HTTPCfg.DebugMode {
		gin.SetMode(gin.ReleaseMode)
	}

	engine := gin.Default()
	pprof.Register(engine)

	s.engine = engine

	// Override Swagger BasePath for reverse proxy deployments
	if bp := os.Getenv("SWAGGER_BASE_PATH"); bp != "" {
		docs.SwaggerInfo.BasePath = bp
	}
        ...
}
```

or add it to backup.yaml config to be consistent with other fields.

If this sounds good, i'd like to make a contribution. 
Thanks.

huanghaoyuanhhy:
Sounds reasonable to me. Would you like to submit a PR for this?

PR #990, opened 2026-03-19
Support customizing Swagger BasePath for reverse proxy 

noisrucer:
Closes: #970 

As discussed in the issue, if the backup is behind a reverse proxy  (e.g., `/{namespace}/backup/`), the swagger does not recognize the proxy path.

The proxy strips the prefix before forwarding to the backup server, so the server has no knowledge of its external path. Since `SwaggerInfo.BasePath` is hardcoded to `/api/v1`, Swagger UI constructs request URLs as `http://{host}/api/v1/restore` instead of `http://{host}/{namespace}/backup/api/v1/restore`.

A simple approach would be to read an env var to override base path at startup.


noisrucer:
@huanghaoyuanhhy
I've updated to use config for consistency and removed the stale `simpleResponse` from README.md and synced current fields.

Thanks

PR #994, opened 2026-03-23
backup: strip segment info from collection and partition level meta

huanghaoyuanhhy:
## Summary
- Strip nested segment data from `collection_meta.json` and `partition_meta.json` to reduce memory usage during backup metadata serialization
- This fixes OOM on large backups where segment binlog info was being serialized 3 extra times beyond `full_meta.json`

## Changes
- `buildCollectionMeta()`: manually construct collection objects without `PartitionBackups` and `L0Segments`, instead of directly serializing the full nested tree
- `buildPartitionMeta()`: manually construct partition objects without `SegmentBackups`
- Add unit tests verifying collection/partition meta are stripped while segment meta and full meta remain complete

## Background
The redundancy was introduced in aaadb04 (Support L0 segments, 2024-06) which changed collection cloning from manual field construction to `proto.Clone`, inadvertently carrying the full partition→segment subtree into `collection_meta.json`. This is backward-compatible because:
1. New backups always have `full_meta.json`, which the read path prefers
2. The level-file fallback `levelToTree()` rebuilds segments from `segment_meta.json` and overwrites whatever is in collection/partition meta
3. L0 segments (2024-06) post-date `full_meta.json` (2023-07), so any backup with L0 segments also has `full_meta.json`

### DIFF
--- a/core/backup/meta_builder.go
+++ b/core/backup/meta_builder.go
@@ -196,9 +196,23 @@ func (builder *metaBuilder) buildCollectionMeta() ([]byte, error) {
 	builder.mu.Lock()
 	defer builder.mu.Unlock()
 
-	// TODO: don't know why we need segment info in collection meta. maybe just a bug.
-	// we should remove it in the future.
-	info := &backuppb.CollectionLevelBackupInfo{Infos: builder.data.GetCollectionBackups()}
+	// Shallow-copy each collection, keep partition metadata but strip all segment data.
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/lgtm

alohaha22 (approved):

alohaha22 (approved):
/lgtm

PR #999, opened 2026-03-27
ci: pin milvus master image to pre-#48477 version

huanghaoyuanhhy:
## Summary
- Temporarily pin the Milvus master image in CI to `master-20260322-aed7c8bc` (the last working image before milvus-io/milvus#48005)
- milvus-io/milvus#48477 introduced an `insert_log` path regression that causes backup CI to fail with `segment has no insert logs`

## Changes
- Replace all `master-latest` references in `main.yaml` matrix and condition checks with the pinned tag
- Add env var `MILVUS_MASTER_IMAGE_TAG` and comments for tracking

Will revert to `master-latest` once the upstream fix lands.

/kind improvement

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -14,6 +14,9 @@ on:
 
 env:
   go-version: 1.25
+  # Pinned to last working master image before milvus-io/milvus#48477 (PR #48005 broke insert_log path).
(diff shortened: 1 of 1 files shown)


huanghaoyuanhhy:
The CI required checks won't report as passed since the matrix values changed and the check names no longer match the branch protection rules. Merging manually.

PR #1011, opened 2026-04-03
ci: add etcd schema verification for secondary restore

huanghaoyuanhhy:
## Summary
- Add standalone tool `cmd/etcd-schema-verify` that reads collection metadata directly from etcd and compares source vs target schemas at the etcd level
- Covers fields invisible to DescribeCollection: internal field IDs, field states, index params, partition states, function schemas
- CDC scenario aware: compares all IDs (collection/field/partition/index) since they are preserved across clusters
- Integrate into secondary restore CI to verify schema alignment after restore

## Changes
- `cmd/etcd-schema-verify/` — standalone Go binary with flag-based CLI
  - Reads databases, collections, fields, partitions, indexes, functions from etcd via protobuf deserialization
  - Diff logic matches by ID, outputs JSON with `aligned` flag and detailed diffs
  - Exit code 1 when not aligned (CI friendly)
- `.github/workflows/main.yaml` — build the tool and run it after secondary restore for `hello_milvus` and `hello_milvus2`

/kind improvement

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -601,6 +601,7 @@ jobs:
           go get
           go build
           go build -o inittarget ./cmd/inittarget
+          go build -o etcd-schema-verify ./cmd/etcd-schema-verify
 
       - name: Install dependency
         timeout-minutes: 5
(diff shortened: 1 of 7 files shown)


wenhuiZilliz (approved):
/lgtm

PR #1014, opened 2026-04-07
ci: add cross-version matrix to repro $meta misalignment

huanghaoyuanhhy:
## Summary

Reproduce the `$meta` field schema misalignment from #1013 in CI by adding a dynamic-schema collection and a new matrix scenario that prepares data on a pre-#46419 Milvus before upgrading to a CDC-capable image for backup/restore.

This is a TDD red test — the new `cross-version-upgrade` matrix entry is expected to fail on `etcd-schema-verify` until the underlying bug is fixed.

## Changes

- `example/prepare_data.py`: add `hello_milvus_dynamic` collection (`enable_dynamic_field=True`, 3000 rows, dynamic fields populated on every row)
- `example/secondary/verify_data.py`: include the new collection in the post-restore verification list
- `.github/workflows/main.yaml` (`test-backup-restore-secondary`):
  - Replace single `image_tag` matrix with `prepare_image_tag` / `run_image_tag` / `scenario` `include` matrix
  - Add `cross-version-upgrade` scenario: prepare on `v2.6.7`, upgrade to `2.6-latest` before backup/restore (CDC support landed in v2.6.11+; volumes are bind-mounted so the pre-#46419 `$meta` schema survives the upgrade)
  - Add new `Upgrade dual Milvus to run image` step, conditional on `prepare != run`
  - Wire `hello_milvus_dynamic` into the index/load step and the `etcd-schema-verify` loop

## Expected CI behavior

| scenario | bug not fixed | bug fixed |
|---|---|---|
| `same-version` (master → master) | pass — both sides have `Nullable=true, DefaultValue={}`, matches the hardcoded restore values | pass |
| `cross-version-upgrade` (v2.6.7 → 2.6-latest) | **fail** — source `$meta` (created on v2.6.7) has no `Nullable` / `DefaultValue`; secondary restore reconstructs it with the hardcoded values, etcd-schema-verify reports a diff | pass |

Related to #1013

/kind improvement

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -572,12 +572,28 @@ jobs:
 
   test-backup-restore-secondary:
     needs: unit-test-go
-    name: Backup and restore secondary
+    name: Backup and restore secondary (${{ matrix.scenario }})
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
-        image_tag: [master-latest]
+        # Two scenarios:
+        #   - same-version:          source and target are both the latest master.
+        #   - cross-version-upgrade: source data is created on a pre-#46419 Milvus
+        #     (v2.6.7, where the auto-appended $meta field has no Nullable nor
+        #     DefaultValue), then the cluster is upgraded to a CDC-capable image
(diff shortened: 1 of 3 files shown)


lentitude2tk (approved):
/lgtm

PR #1015, opened 2026-04-07
milvus: use strict semver to parse server version for feature check

huanghaoyuanhhy:
## Summary
The lenient `semver.NewVersion` parses Milvus dev RC build tags like `2.6-20260404-31fb3fc` as `2.6.0-20260404-31fb3fc`, a prerelease of `2.6.0`. By semver precedence rules, a version with a prerelease suffix compares **less** than the same `MAJOR.MINOR.PATCH` without one, so this dev tag ends up below `2.6.5-0`, `2.6.8-0` and `2.6.11-0`. As a result, `MultiL0InOneJob`, `FlushAll`, `CollectionLevelGCControl` and `FuncRuntimeCheck` are silently disabled when running against a tip-of-2.6 dev build.

Switch `checkFeature` to `semver.StrictNewVersion`, which requires `MAJOR.MINOR.PATCH`. Non-strict tags (`2.6-20260404-31fb3fc`, `master-20260226-abcdef`, `""`) now fall through to `_latestDevVersion` and pass all `>= X` constraints, restoring the intended behavior.

The early-return in `GetVersion` keeps the lenient parser on purpose: it is only a "skip extra RPC" sanity check, not a constraint comparison, so widening it to strict would just cost an extra RPC round-trip with no behavioral benefit.

## Changes
- Extract `parseServerVersionForFeature` helper that uses `semver.StrictNewVersion` and falls back to `_latestDevVersion` on failure
- Use the helper from `checkFeature`
- Add `TestParseServerVersionForFeature` covering release versions (`2.6.0` / `2.6.5` / `2.6.10` / `2.6.11`), the `2.6-20260404-31fb3fc` dev tag, master tags, and the empty string
- Update the `_latestDevVersion` doc comment to mention the `2.6-…` tag form

/kind improvement

### DIFF
--- a/internal/client/milvus/grpc.go
+++ b/internal/client/milvus/grpc.go
@@ -57,9 +57,10 @@ type featureTuple struct {
 	Flag        FeatureFlag
 }
 
-// _latestDevVersion is used as a fallback when the server returns a non-semver version
-// string (e.g. "master-20260226-abcdef" from dev builds). It ensures lower-bound
-// constraints (>= X) pass while upper-bound constraints (< Y) correctly fail.
+// _latestDevVersion is used as a fallback when the server returns a version string
+// that is not a strict MAJOR.MINOR.PATCH semver (e.g. "master-20260226-abcdef" or
+// "2.6-20260404-31fb3fc" from dev builds). It ensures lower-bound constraints (>= X)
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/lgtm

PR #1017, opened 2026-04-07
backup: capture dynamic field schema from etcd for secondary restore

huanghaoyuanhhy:
## Summary
Capture the `$meta` dynamic field schema directly from etcd at backup time, since the Milvus `DescribeCollection` API filters out `IsDynamic` fields and leaves the gRPC-based backup path with no way to see the actual attributes (`Nullable`, `DefaultValue`, ...). Without this, secondary restore had to reconstruct the field with hardcoded defaults that did not match the source collection and broke its strict alignment requirement.

## Changes
- Add `core/backup/coll_dyn_field_task.go` that reads field schemas from etcd at `{rootPath}/meta/root-coord/fields/{collectionID}/{fieldID}`, filters for `IsDynamic == true`, and groups them by collection, mirroring the existing `coll_index_extra_task.go` pattern.
- Wire `backupCollDynField` into `Task.privateExecute`, reusing the etcd client created for `BackupIndexExtra`.
- Add `metaBuilder.addDynamicFields` to inject the dynamic field into each `CollectionBackupInfo.Schema.Fields`, plus a `convDynField` helper for the schemapb→backuppb conversion.
- Drop any dynamic field from the regular restore's `fields()` split, because `CreateCollection` auto-creates `$meta` and `AddField` rejects dynamic fields.
- Remove the legacy `appendDynamicField` from secondary restore (which synthesized hardcoded `Nullable=true` / `DefaultValue={}`) and trust the backup file. Add `checkDynamicField` to fail fast when `EnableDynamicField=true` but the backup metadata does not contain `$meta` (i.e. an old backup taken before this fix); the error message points users to re-run backup with `BackupIndexExtra` enabled.
- Unit tests for `parseCollIDFromFieldKey`, `metaBuilder.addDynamicFields`, and `checkDynamicField`.

Fixes #1013

### DIFF
--- a/core/backup/coll_dyn_field_task.go
+++ b/core/backup/coll_dyn_field_task.go
@@ -1,0 +1,115 @@
+package backup
+
+import (
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/milvus-io/milvus-proto/go-api/v2/schemapb"
+	clientv3 "go.etcd.io/etcd/client/v3"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/zilliztech/milvus-backup/internal/log"
+)
+
+// collDynFieldTask reads the dynamic field ($meta) schema for each collection
+// directly from etcd. The Milvus DescribeCollection API filters out fields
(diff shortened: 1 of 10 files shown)


lentitude2tk (approved):
/lgtm

PR #1018, opened 2026-04-08
ci: cover add_collection_field path in secondary restore test

huanghaoyuanhhy:
## Summary
Extend the secondary restore CI fixtures to exercise the `add_collection_field` schema path, on top of the existing dynamic-field coverage from #1014.

## Changes
- `example/prepare_data.py`: add a `hello_milvus_added` collection that inserts half its rows, flushes, then uses `MilvusClient.add_collection_field` to add five nullable fields (`added_int64`, `added_varchar`, `added_double`, `added_bool`, `added_json`, four of them with `default_value`), then inserts the remaining rows populating the new fields and flushes again.
- `example/secondary/verify_data.py`: include `hello_milvus_added` in the post-restore verification list.
- `.github/workflows/main.yaml`: include `hello_milvus_added` in the upstream `create_index/load` step and in the `etcd-schema-verify` loop, so source/target schemas for the added fields are compared.

The added fields carry `Nullable=true` and `default_value` attributes that must round-trip through backup metadata into secondary restore. If the backup tool drops or hardcodes these attributes the source/target schemas will diverge in `etcd-schema-verify`, exposing the same class of schema misalignment as the dynamic-field `$meta` path.

Related to #1013

/kind improvement

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -673,6 +673,7 @@ jobs:
         shell: bash
         run: |
           python example/prepare_data.py --uri http://127.0.0.1:19530
+          python example/prepare_data.py --uri http://127.0.0.1:19530 --scenario added
 
       - name: Upgrade dual Milvus to run image
         if: ${{ matrix.prepare_image_tag != matrix.run_image_tag }}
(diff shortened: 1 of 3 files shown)


lentitude2tk (approved):
/lgtm

PR #1019, opened 2026-04-08
test: verify rbac grant content after restore, not just names

huanghaoyuanhhy:
## Summary
The existing RBAC backup/restore E2E (`test_milvus_restore_backup_with_rbac`) only asserts that user / role / privilege-group **names** exist after restore. It never inspects the restored grant content, the privilege-group payload, or the user-role binding, so a regression that silently dropped or mangled any of those would still pass CI.

## Changes
- Add three reusable helpers on `TestRestoreBackupWithRbac`:
  - `verify_role_grants` — match grants by `object_name` / `privilege` / `db_name` (and optionally `object_type`); only the keys present in the expected dict are compared, so callers can omit fields whose value is server-inferred.
  - `verify_privilege_group_content` — assert a privilege group exists and contains the expected privileges.
  - `verify_user_roles` — assert a user is bound to the expected roles via `describe_user`.
- Wire all three into the existing `test_milvus_restore_backup_with_rbac` after restore, so we now check that the wildcard grant (`object_name='*'`, `db_name='*'`), the `Insert` privilege inside the custom group, and the user→role binding all survive backup → drop → restore intact.

/kind improvement

### DIFF
--- a/tests/testcases/test_restore_backup_with_rbac.py
+++ b/tests/testcases/test_restore_backup_with_rbac.py
@@ -103,6 +103,63 @@ class TestRestoreBackupWithRbac(TestcaseBase):
             assert group in privilege_list, \
                 f"Privilege group {group} not restored"
 
+    @staticmethod
+    def verify_role_grants(client, role_name, expected_grants):
+        """Verify a role has the expected grants restored.
+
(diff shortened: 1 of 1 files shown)


lentitude2tk (approved):
/lgtm

PR #1020, opened 2026-04-08
stream: move time-tick allocation into stream client

huanghaoyuanhhy:
## Summary
Sink wire time-tick allocation into `milvus.StreamClient` and change the `Stream` interface to take `MutableMessage`. The stream client now owns the monotonic counter and serializes (alloc, dispatch) so per-pchannel time-tick monotonicity holds across concurrent restore goroutines, removing the need to thread `tsAlloc` / `sendMu` through every secondary task.

This is the first step of a larger reliability improvement for the secondary restore replicate path; a follow-up PR will add an in-memory replay buffer keyed by time-tick on top of this interface.

## Changes
- `Stream` interface now exposes `Alloc()`, `Send(MutableMessage)`, `Forward(ImmutableMessage)`, `WaitConfirm()`
- `StreamClient` owns `tsAlloc` and a `dispatchMu` that atomically pairs (alloc tt, mutable→immutable conversion, dispatch to pchClient)
- `Forward` is added for replaying pre-built immutable messages whose time-ticks come from the source cluster (flush-all messages); they are not re-stamped on the target side
- `Alloc` is exposed so callers keep `MsgBase.Timestamp` on the same monotonic axis as the wire time-tick (verified with the Milvus broker that the body Timestamp is inert on the receive path, only the wire time-tick is consumed)
- `tsAlloc` and `fakeMessageID` moved from `core/restore/secondary` into `internal/client/milvus`
- All 6 producer call sites in `core/restore/secondary` simplified to pass mutable messages; `tsAlloc` / `sendMu` plumbing deleted from `Task`, `databaseTask`, `collDDLTask`, `collDMLTask`, `collLoadTask` and their args structs
- `Task.sendFlushAll` switched to `Stream.Forward` to preserve source-cluster time-ticks
- `recordingStream` test mock rewritten to mimic the production client (internal alloc + immutable conversion) so existing per-pch monotonicity assertions still surface

/kind improvement


lentitude2tk (approved):
/LGTM

lentitude2tk (approved):
/lgtm

PR #1021, opened 2026-04-08
test: add regression for restoring wildcard '*' privilege grants

huanghaoyuanhhy:
## Summary
Add an E2E regression test that exercises the path where a grant's privilege name is `'*'` (the AnyWord, meaning "any privilege"). The Milvus admin role uses this internally, and users may grant it to custom roles too. The Milvus server's `OperatePrivilege` path handles `'*'` correctly, but the `RestoreRBAC` path historically did not — producing either a `privilege group [] does not exist` error on 2.4.x / 2.5.x or silently storing the grant as `PrivilegeGroup*` on master.

The existing `test_milvus_restore_backup_with_rbac` does not catch this for two compounding reasons:
1. It only grants a custom *privilege group* on `collection_name='*'` / `db_name='*'`, never the privilege name `'*'` itself, so the broken `kv_catalog.RestoreRBAC` branch is never hit.
2. It runs source and destination on the same cluster and never drops `admin`, so even if the backup contained admin's `'*'` grant, `conv.Grants` would dedup it against the still-existing one in the target and never send it to the server.

This new test grants `'*'` directly on a custom role and then drops the role before restore, defeating the dedup so the broken path actually executes.

## Changes
- Add `test_milvus_restore_backup_with_wildcard_privilege` in `tests/testcases/test_restore_backup_with_rbac.py`:
  - Create a custom user/role and call `grant_privilege_v2(role, privilege='*', collection_name='*', db_name='*')`.
  - Sanity-check that the source cluster accepted the grant (turns any future server-side regression in `OperatePrivilege` into a clear setup-time failure rather than a confusing restore-time error).
  - Backup with `rbac=True`, then `clean_rbac` to drop the custom role/user (so dedup in `conv.Grants` cannot mask the broken path), then restore.
  - Assert via `verify_role_grants` that the grant comes back with `object_name='*'`, `privilege='*'`, `db_name='*'`. This catches both failure modes: a loud `privilege group [] does not exist` on the restore call, and the silent `PrivilegeGroup*` corruption on master.
- Reuses the helpers added in #1019 (`verify_role_grants`, `verify_user_roles`).

/kind improvement

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -1057,7 +1057,11 @@ jobs:
         shell: bash
         working-directory: tests
         run: |
-          pytest -s -v --tags ${{ matrix.case_tag }} -n 4 --tei_endpoint http://text-embeddings:80
+          if [ ${{ matrix.case_tag }} == "RBAC" ]; then
+            pytest -s -v --tags ${{ matrix.case_tag }}
+          else
+            pytest -s -v --tags ${{ matrix.case_tag }} -n 4 --tei_endpoint http://text-embeddings:80
+          fi
 
       - name: Get Milvus status
         shell: bash
--- a/.github/workflows/nightly.yaml
+++ b/.github/workflows/nightly.yaml
@@ -89,7 +89,9 @@ jobs:
         shell: bash
         working-directory: tests
         run: |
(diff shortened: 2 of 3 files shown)


lentitude2tk (approved):
LGTM

PR #1022, opened 2026-04-08
stream: add per-pchannel replay buffer for replicate stream

huanghaoyuanhhy:
## Summary
Add a bounded in-memory replay buffer per pchannel inside the replicate stream client. The previous implementation wrote each message directly onto the gRPC bidi stream and exited \`recvLoop\` on the first error, silently losing in-flight messages on any connection drop and deadlocking \`WaitConfirm\` once recv had quit. The new implementation matches the pattern Milvus' own CDC client uses internally (\`internal/cdc/replication/replicatestream\`).

This is the second part of the secondary-restore reliability work that started in #1020.

## Changes
- Add a bounded \`msgQueue\` (capacity 4096, blocks producers on backpressure) with two cursors: a read cursor that \`sendLoop\` advances as it ships messages, and a storage head that only moves on broker confirms via \`CleanupConfirmedMessages\`. \`SeekToHead\` rewinds the read cursor on reconnect so unconfirmed messages are replayed in time-tick order on the new gRPC stream.
- Restructure \`pchClient\` as three cooperating goroutines:
  - \`runForever\` — outer reconnect loop with exponential backoff (100ms → 10s)
  - \`sendLoop\` — drains the queue onto the current stream
  - \`recvLoop\` — reads \`ConfirmedTimeTick\` and drops the matching prefix
- \`Send\`/\`Forward\` enqueue into the queue and return as soon as the message is buffered, decoupling producers from network latency and reconnects.
- \`WaitConfirm\` now takes a \`context.Context\` and returns an \`error\` so callers can bail on cancellation. \`secondary.Task.Execute\` propagates the context and surfaces the error; \`Task.closeClients\` calls the new \`Stream.Close\` so per-pch goroutines are bounded by the restore task.
- The \`recordingStream\` test fake in \`core/restore/secondary\` is updated for the new interface.
- Unit tests cover the queue (enqueue/cleanup/seek/cursor adjustment/backpressure/ctx cancellation/close) and the pchClient (confirm drains buffer, partial confirm drops prefix, reconnect replays unconfirmed in tt order, new sends after reconnect flow to the new stream, WaitConfirm respects ctx cancel) using a hand-rolled \`fakeReplicateStream\` that mirrors how a real gRPC bidi stream tears down on context cancellation.

## Notes on duplicate-on-replay
The Milvus replicate interceptor dedupes per checkpoint, not per message: messages with \`tt > checkpoint\` are reprocessed if replayed (\`internal/streamingnode/server/wal/interceptors/replicate/replicates/impl.go\`). Milvus' own CDC client makes the same trade-off — accepts the rare double-execution risk in exchange for not losing data on every reconnect — and we follow suit.

/kind improvement

### DIFF
--- a/core/restore/secondary/coll_dml_task_test.go
+++ b/core/restore/secondary/coll_dml_task_test.go
@@ -58,7 +58,9 @@ func (s *recordingStream) Forward(_ context.Context, msg *commonpb.ImmutableMess
 	return nil
 }
 
-func (s *recordingStream) WaitConfirm() {}
+func (s *recordingStream) WaitConfirm(_ context.Context) error { return nil }
+
+func (s *recordingStream) Close() {}
 
 // timestampsByPch returns the time-ticks grouped by physical channel, in send
 // order.
--- a/core/restore/secondary/task.go
+++ b/core/restore/secondary/task.go
@@ -89,6 +89,9 @@ func (t *Task) initClients() error {
 }
 
 func (t *Task) closeClients() {
+	if t.streamCli != nil {
+		t.streamCli.Close()
+	}
 	if t.grpc != nil {
 		if err := t.grpc.Close(); err != nil {
 			t.logger.Warn("close grpc client", zap.Error(err))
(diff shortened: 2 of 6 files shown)


lentitude2tk (approved):
/lgtm

PR #1028, opened 2026-04-17
stream: add in-memory replay queue for replicate stream

huanghaoyuanhhy:
## Summary

Each replicate `pchClient` now owns a per-pchannel in-memory replay buffer (`msgQueue`). `Send`/`Forward` enqueue into the buffer; a background sender drains it onto the current gRPC stream and keeps every message around until the broker confirms its time-tick. On reconnect the read cursor is rewound so any still-unconfirmed message is replayed in time-tick order on the fresh stream.

This fixes two silent data-loss paths in the previous implementation:
1. `cli.Send` only enqueues into the local gRPC transport — if the stream broke before the broker received the message, it was gone forever and its confirm never came.
2. `recvLoop` exited on the first `Recv` error and was never restarted, so later confirms were dropped and `WaitConfirm` deadlocked.

## Changes

- New `msgQueue` interface + `memMsgQueue` impl with two cursors: a storage head advanced only by `Confirm`, and a read cursor rewound by `SeekToHead` on reconnect. Full unit-test coverage for enqueue/read, ctx cancel, `Confirm` prefix drop + cursor adjust, `SeekToHead` replay, `WaitEmpty`, and a concurrent producer/consumer stress test.
- `pchClient` restructured into three cooperating goroutines: `runForever` (reconnect loop with exponential backoff), `sendLoop` (drains queue), `recvLoop` (applies `ConfirmedTimeTick` acks).
- `Stream` interface reshaped:
  - `Send(ctx, build func(ts uint64) []MutableMessage)` — callback style. `build` runs under the dispatch lock with a freshly allocated wire time-tick; callers may use `ts` for body timestamps so body and wire time-ticks always match (required by the broker's import subsystem). Every returned message is stamped with the same `ts` and enqueued. Atomicity gives per-pchannel monotonic wire ticks without any external locking.
  - `Forward(ctx, msgs...)` — enqueues pre-built immutable messages (flush-all replay) whose time-ticks come from the source cluster.
  - `Close` stops every per-pch reconnect loop — wired into `Task.closeClients` so the long-running server path no longer leaks goroutines.
- All `secondary/*` task files drop their `tsAlloc`/`sendMu` plumbing; the callback API consolidates `(alloc ts + build body + stamp wire ts + enqueue)` into one atomic block. `recordingStream` test fake updated accordingly.
- `FakeMessageID` moved into the `milvus` package and exported so `Send` can build immutables internally.

The replay relies on the broker's per-checkpoint dedup, which silently skips any time-tick at or below the checkpoint — the same trade-off Milvus' own CDC client makes.

/kind improvement


lentitude2tk (approved):
LGTM

PR #1031, opened 2026-04-17
storage: Reduce BucketExist probe latency on large buckets by stopping the ListObjects scan after the first result while preserving caller context cancellation.

Sallery-X:


huanghaoyuanhhy:
Thanks for the fix — the `MaxKeys` analysis is spot-on, the high-level `ListObjects` really does paginate per-page regardless of `MaxKeys`.

That said, this change has a regression: `Core.ListObjectsV2` is the no-context variant of the SDK — its implementation internally uses `context.Background()`. So `_ = ctx` silently drops the caller's context, which breaks timeout/cancellation propagation for `CreateBucketIfNotExist` and similar callers.

I'd suggest staying on the high-level `ListObjects` and returning after the first object, with a cancelable sub-context to stop the background goroutine:

```go
subCtx, cancel := context.WithCancel(ctx)
defer cancel()
for obj := range m.cli.ListObjects(subCtx, m.cfg.Bucket, minio.ListObjectsOptions{Prefix: prefix, MaxKeys: 1}) {
    if obj.Err != nil {
        if minio.ToErrorResponse(obj.Err).Code == "NoSuchBucket" {
            return false, nil
        }
        return false, fmt.Errorf("storage: %s list objects %w", m.cfg.Provider, obj.Err)
    }
    return true, nil
}
return true, nil
```

`defer cancel()` makes the SDK's goroutine exit via its `case <-ctx.Done()` branch as soon as we return, so only a single HTTP request is issued and the caller's context still propagates.

huanghaoyuanhhy:
One more option worth mentioning: `ListObjectsIter` (introduced in minio-go v7.0.92) would also work here — it's the Go 1.23 range-over-func variant with no background goroutine, so there's no channel-leak concern at all. Something like:

```go
subCtx, cancel := context.WithCancel(ctx)
defer cancel()
for obj := range m.cli.ListObjectsIter(subCtx, m.cfg.Bucket, minio.ListObjectsOptions{Prefix: prefix, MaxKeys: 1}) {
    if obj.Err != nil {
        if minio.ToErrorResponse(obj.Err).Code == "NoSuchBucket" {
            return false, nil
        }
        return false, fmt.Errorf("storage: %s list objects %w", m.cfg.Provider, obj.Err)
    }
    return true, nil
}
return true, nil
```

Note: the `subCtx, cancel := context.WithCancel(ctx)` + `defer cancel()` is still needed here, because after the first yield succeeds the iterator would otherwise loop back and fire a second `listObjectsV2Query` before the next yield returns false. Canceling the ctx hits the iterator's `contextCanceled(ctx)` check at the top of the outer loop and stops that second request.

Either approach (channel-based or iter-based) works — feel free to pick whichever feels cleaner.

Sallery-X:
> Thanks for the fix — the `MaxKeys` analysis is spot-on, the high-level `ListObjects` really does paginate per-page regardless of `MaxKeys`.
> 
> That said, this change has a regression: `Core.ListObjectsV2` is the no-context variant of the SDK — its implementation internally uses `context.Background()`. So `_ = ctx` silently drops the caller's context, which breaks timeout/cancellation propagation for `CreateBucketIfNotExist` and similar callers.
> 
> I'd suggest staying on the high-level `ListObjects` and returning after the first object, with a cancelable sub-context to stop the background goroutine:
> 
> ```go
> subCtx, cancel := context.WithCancel(ctx)
> defer cancel()
> for obj := range m.cli.ListObjects(subCtx, m.cfg.Bucket, minio.ListObjectsOptions{Prefix: prefix, MaxKeys: 1}) {
>     if obj.Err != nil {
>         if minio.ToErrorResponse(obj.Err).Code == "NoSuchBucket" {
>             return false, nil
>         }
>         return false, fmt.Errorf("storage: %s list objects %w", m.cfg.Provider, obj.Err)
>     }
>     return true, nil
> }
> return true, nil
> ```
> 
> `defer cancel()` makes the SDK's goroutine exit via its `case <-ctx.Done()` branch as soon as we return, so only a single HTTP request is issued and the caller's context still propagates.

ok

PR #1043, opened 2026-05-09
Fix restore import paths to avoid empty S3 prefix listing

Sallery-X:
This is triggered by insert-only restore data: the backup has insert logs, but no delta logs, so `insertLogDir` is set and `deltaLogDir` is empty.

The affected paths are regular RESTful restore import and secondary restore import-file construction. Both now filter empty  paths so Milvus will not receive `""` and list S3 with `prefix=`.

gRPC bulk insert is intentionally kept unchanged via `toGrpcPaths`. The old gRPC import API expects paths in positional `[insert, delta]` form, so for insert-only data `["<insert_log_dir>", ""]` is the compatibility format meaning “insert logs exist, delta logs do not”.

### DIFF
--- a/core/restore/coll_task.go
+++ b/core/restore/coll_task.go
@@ -391,6 +391,17 @@ func (ct *collTask) restoreNotL0SegV1(ctx context.Context, part *backuppb.Partit
 }
 
 func toPaths(dir partitionDir) []string {
+	paths := make([]string, 0, 2)
+	if dir.insertLogDir != "" {
+		paths = append(paths, dir.insertLogDir)
+	}
+	if dir.deltaLogDir != "" {
+		paths = append(paths, dir.deltaLogDir)
+	}
+	return paths
+}
+
+func toGrpcPaths(dir partitionDir) []string {
 	if len(dir.insertLogDir) == 0 {
 		return []string{dir.deltaLogDir}
 	}
@@ -741,13 +752,13 @@ func (ct *collTask) bulkInsertViaGrpc(ctx context.Context, partitionName string,
 		g.Go(func() error {
 			defer ct.bulkInsertSem.Release(1)
 
-			paths := toPaths(dir)
+			paths := toGrpcPaths(dir)
 			ct.logger.Info("start bulk insert via grpc", zap.Strings("paths", paths), zap.String("partition", partitionName))
 			in := milvus.GrpcBulkInsertInput{
 				DB:             ct.targetNS.DBName(),
 				CollectionName: ct.targetNS.CollName(),
 				PartitionName:  partitionName,
-				Paths:          toPaths(dir),
+				Paths:          paths,
 				BackupTS:       b.timestamp,
 				IsL0:           b.isL0,
 				StorageVersion: b.storageVersion,
--- a/core/restore/coll_task_test.go
+++ b/core/restore/coll_task_test.go
@@ -83,11 +83,33 @@ func TestToPaths(t *testing.T) {
 	// without delta
 	dir = partitionDir{insertLogDir: "insert"}
 	paths = toPaths(dir)
-	assert.Equal(t, []string{"insert", ""}, paths)
+	assert.Equal(t, []string{"insert"}, paths)
 
 	// without insert
 	dir = partitionDir{deltaLogDir: "delta"}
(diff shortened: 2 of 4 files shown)


huanghaoyuanhhy:
Hi @Sallery-X, could you add a bit more context to the PR description?

From the diff, this seems to filter empty restore import paths so S3 listing will not accidentally use an empty prefix, while keeping the old `[insert, delta]` shape for gRPC bulk insert via `toGrpcPaths`.

Could you clarify:
1. What concrete restore scenario triggers the empty-prefix listing issue? For example, insert-only backup data, delta-only data, or an empty partition?
2. Which restore path was affected: regular import, secondary restore, gRPC bulk insert, or all of them?
3. Why does gRPC bulk insert still need to keep an empty delta path when insert logs exist but delta logs do not?

This would make the behavioral intent easier to review and help confirm the compatibility expectation.

Sallery-X:
> Hi @Sallery-X, could you add a bit more context to the PR description?
> 
> From the diff, this seems to filter empty restore import paths so S3 listing will not accidentally use an empty prefix, while keeping the old `[insert, delta]` shape for gRPC bulk insert via `toGrpcPaths`.
> 
> Could you clarify:
> 
> 1. What concrete restore scenario triggers the empty-prefix listing issue? For example, insert-only backup data, delta-only data, or an empty partition?
> 2. Which restore path was affected: regular import, secondary restore, gRPC bulk insert, or all of them?
> 3. Why does gRPC bulk insert still need to keep an empty delta path when insert logs exist but delta logs do not?
> 
> This would make the behavioral intent easier to review and help confirm the compatibility expectation.

done

huanghaoyuanhhy:
Thanks for the update — the description is clearer now. One thing I'm still not sure about: what's the actual symptom? It says insert-only data triggers it, but CI is green, so before merging I'd like to understand what actually breaks — a real failure somewhere, or something you noticed during review?

Sallery-X:
> Thanks for the update — the description is clearer now. One thing I'm still not sure about: what's the actual symptom? It says insert-only data triggers it, but CI is green, so before merging I'd like to understand what actually breaks — a real failure somewhere, or something you noticed during review?

  The symptom is an import timeout.

  For insert-only data we were sending ["insert_path", ""] in the RESTful/
  secondary import payload. The empty second path makes the import side list the
  bucket with an empty prefix, which can scan a large bucket and time out. Small
  buckets may not hit it, so CI stays green.

  This is why the fix filters empty paths from the RESTful/secondary payload. gRPC
  is kept unchanged because it still depends on the insert/delta offset
  convention.

Sallery-X:
> Thanks for the update — the description is clearer now. One thing I'm still not sure about: what's the actual symptom? It says insert-only data triggers it, but CI is green, so before merging I'd like to understand what actually breaks — a real failure somewhere, or something you noticed during review?

I did not discover this issue during the review process. I found it while performing the backup and restore process.
so please review @huanghaoyuanhhy @zhuwenxing

huanghaoyuanhhy:
Why split `toPaths` and `toGrpcPaths`? Is there a specific code path or error that requires keeping the old form?

huanghaoyuanhhy:
Sorry for the slow review — I've been tied up with internal work recently, and I also wanted to take time to evaluate this carefully: splitting `toPaths` and `toGrpcPaths` adds maintenance and cognitive cost, so I've been checking the Milvus side to confirm whether there's an actual constraint requiring the old form.

Sallery-X:
> Sorry for the slow review — I've been tied up with internal work recently, and I also wanted to take time to evaluate this carefully: splitting `toPaths` and `toGrpcPaths` adds maintenance and cognitive cost, so I've been checking the Milvus side to confirm whether there's an actual constraint requiring the old form.

The original issue was caused by the old `toPaths` returning `[insertPath,
  ""]` when the segment has insert logs but no delta logs. That empty path is
  invalid for the REST/secondary import path and can cause Milvus import to fail.

  I split `toPaths` and `toGrpcPaths` because this PR was intended to fix the
  empty-path issue without changing the existing gRPC import payload at the same
  time. The gRPC import path has historically used a positional `[]string` format
  where index 0 is insert log and index 1 is delta log, so I kept that behavior to
  reduce compatibility risk.

PR #1047, opened 2026-05-17
fix(storage): skip custom endpoint for standard GCS in gcpnative client

DashUmEr:
## Problem

The `gcpnative` storage type cannot connect to a normal Google Cloud Storage bucket. Client init fails with:

```
dialing: multiple credential options provided
```

This makes `gcpnative` effectively unusable for the standard GCS service — the very thing it was added for in #625 (which implemented #444).

## Root cause

1. `internal/storage/factory.go` always builds `Config.Endpoint` via `net.JoinHostPort(address, port)`. For a normal GCS backup the endpoint is therefore **never empty** — it is something like `storage.googleapis.com:443`.
2. `newGCPNativeClient` in `internal/storage/gcp_native.go` then does:
   ```go
   if cfg.Endpoint != "" {
       opts = append(opts, option.WithEndpoint("https://"+cfg.Endpoint+"/storage/v1/"))
   }
   ```
   Because `cfg.Endpoint` is never empty, this **always** passes a custom endpoint.
3. For the standard GCS host, passing `option.WithEndpoint` makes `cloud.google.com/go/storage.NewClient` treat it as a *custom* endpoint and inject its own conflicting auth handling. That collides with the explicit `option.WithCredentials(...)` already passed alongside it, and client init fails with `dialing: multiple credential options provided`.

In other words: the `Endpoint != ""` guard was meant to distinguish "standard GCS" from "custom emulator endpoint", but `factory.go` makes the endpoint non-empty in *both* cases, so the guard never fires for the standard case.

## Fix

Only pass `option.WithEndpoint` for a genuinely non-GCS host (an emulator such as `fake-gcs-server`). For the standard GCS service the endpoint is left unset so the native client uses its own defaults.

A new `isStandardGCSEndpoint` helper recognises the public GCS hosts after stripping the `:port` suffix that `net.JoinHostPort` always appends:

- `storage.googleapis.com`
- `*.storage.googleapis.com`
- `www.googleapis.com`
- `googleapis.com`
- empty string

This makes `gcpnative` usable for normal GCS buckets while keeping emulator support intact — a genuinely custom host still gets `WithEndpoint`.

## Scope

This PR touches only `internal/storage/gcp_native.go`. It is a one-function, behaviour-preserving change for emulators and a bug fix for standard GCS.

## Related (not addressed here)

There is a separate latent issue: the pinned `google.golang.org/api v0.258.0` carries a credential-handling bug ([googleapis/google-api-go-client#3414](https://github.com/googleapis/google-api-go-client/issues/3414), fixed in `v0.259.0`). That can surface its own credential errors with `gcpnative`. It is **not** touched by this PR — flagged here only as related context for anyone debugging `gcpnative` auth; a dependency bump would be a separate change.

Refs #444

### DIFF
--- a/internal/storage/gcp_native.go
+++ b/internal/storage/gcp_native.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	"cloud.google.com/go/storage"
 	"go.uber.org/zap"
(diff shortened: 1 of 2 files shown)


DashUmEr:
Friendly nudge for a maintainer review when there's a chance 🙏

This is a small, self-contained fix for a real bug that makes the `gcpnative` storage client unusable against standard GCS: `factory.go` always sets `Config.Endpoint`, and `gcp_native.go` unconditionally turns that into an `option.WithEndpoint(...)`. Combined with ADC / a credentials file, the Go GCS SDK then rejects the client with `multiple credential options provided`.

The patch skips the custom-endpoint option when the endpoint is the standard GCS host, leaving genuinely custom endpoints (test emulators, private GCS) untouched. DCO and CI checks are green.

Happy to add a regression test or adjust the scope if a maintainer would prefer — just let me know.

huanghaoyuanhhy:
Thanks! Yes please add unit tests for `isStandardGCSEndpoint` before merging — there's no `gcp_native_test.go` yet.

DashUmEr:
Added `internal/storage/gcp_native_test.go` in 9a071e5 — a table-driven `TestIsStandardGCSEndpoint`:

- **Standard GCS → `true`** (WithEndpoint skipped): `storage.googleapis.com` with and without port, virtual-hosted bucket sub-hosts (`mybucket.storage.googleapis.com`), `www.googleapis.com`, bare `googleapis.com`, and the empty / `:port`-only edge cases.
- **Custom & emulator → `false`** (WithEndpoint set): `localhost`, `fake-gcs-server`, loopback + IPv6 hosts, custom domains, and a `googleapis.com` lookalike domain (`storage.googleapis.com.evil.example`) to guard the suffix check.

Thanks for the review!

DashUmEr:
Friendly re-nudge 🙏 — the unit tests @huanghaoyuanhhy asked for landed in `9a071e5` (`internal/storage/gcp_native_test.go`, table-driven `TestIsStandardGCSEndpoint`), and CI is green. Whenever a maintainer has a moment for a review/merge it would be much appreciated — happy to rebase if it has drifted. Thanks!

PR #1048, opened 2026-05-19
test: add e2e cases for nullable vector backup and restore

huanghaoyuanhhy:
## Summary
Add end-to-end coverage for nullable vector fields (Milvus 2.6.18 / 3.0.0 feature) through the backup/restore round-trip. Existing nullable tests only cover scalar fields (FLOAT, VARCHAR, JSON, ARRAY) — there is no coverage that the `nullable=true` flag and per-row NULL semantics survive backup + restore for vector fields.

## Changes
- Add `tests/testcases/test_restore_nullable_vector.py` with four test cases:
  - `test_restore_nullable_vector_round_trip` parameterized over the six vector dtypes in scope (FLOAT, FLOAT16, BFLOAT16, INT8, BINARY, SPARSE_FLOAT): asserts the restored field keeps `nullable=true` and that NULL / non-NULL rows are preserved per primary key.
  - `test_restore_nullable_vector_search_skips_null`: builds an index on the restored collection and confirms search hits are at most the non-NULL row count and never land on a NULL row.
  - `test_restore_add_nullable_vector_field`: exercises `add_collection_field` with a nullable vector field, then checks the field round-trips and pre-add rows come back as NULL.
  - `test_restore_nullable_vector_with_skip_create_collection`: negative case asserting that restoring NULL vector rows into a pre-created target with `nullable=false` fails fast (L2; expected to harden further once the restore path adds a pre-flight nullable schema check).

The tests bypass `compare_collections` and inspect rows directly via `MilvusClient.query`, because `compare_collections` does a strict `src_res[i] == dist_res[i]` which is ambiguous for numpy ndarray vector values and may not handle nullable vectors cleanly.

/kind improvement

### DIFF
--- a/tests/testcases/test_restore_nullable_vector.py
+++ b/tests/testcases/test_restore_nullable_vector.py
@@ -1,0 +1,426 @@
+"""Backup/restore round-trip coverage for nullable vector fields.
+
+PRD reference: Zilliz Cloud Nullable Vector (Milvus 2.6.18 / 3.0.0).
+These tests assert that:
+  1. The `nullable=True` flag on vector fields survives backup -> restore.
+  2. NULL vector values round-trip as NULL (not zero vectors, not omitted rows).
+  3. Mixed NULL / non-NULL data preserves per-row vector content.
+  4. AddCollectionField of a nullable vector field is captured by backup and
+     restored as a nullable field on the target collection.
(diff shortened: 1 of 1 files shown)


lentitude2tk (approved):
/LGTM

PR #1052, opened 2026-05-21
storage: replace per-file copy verification with batched prefix verify

### LINKED ISSUE #1050: [Bug]: migrate fails when stage credentials allow PutObject but deny HeadObject
## Summary

`milvus-backup migrate` fails after successfully applying a Zilliz Cloud Stage because it verifies each uploaded object with `HeadObject`. In our UAT3 BYOC migration test, the temporary Stage credentials allow `PutObject`, but `HeadObject` returns `AccessDenied`, so the CLI exits before it can call `/v2/migrations/startByRemote`.

Manual migration succeeds when using the same backup and Stage but skipping the client-side `HeadObject` verification and directly calling `startByRemote` after upload.

## Environment

- milvus-backup: `v0.5.14`
- Source Milvus: `2.6.16` standalone, Docker Compose + MinIO
- Target: Zilliz Cloud UAT3 BYOC-I cluster
- Migration mode: `milvus-backup migrate`
- Backup: created by `milvus-backup create`, contains one collection

## Reproduction

1. Create a local Milvus collection and insert test data.
2. Create a backup:

```bash
milvus-backup \
  --config configs-uat3-local.yaml \
  create \
  -n regeneron_round2_20260520_backup \
  --filter default.regeneron_restore_src_round2_20260520 \
  --strategy serial_flush
```

3. Run migration to a Zilliz Cloud BYOC target cluster:

```bash
milvus-backup \
  --config configs-uat3-local-with-api-key.yaml \
  migrate \
  -c <target-cluster-id> \
  -n regeneron_round2_20260520_backup
```

## Actual behavior

Stage apply succeeds:

```text
[migrate/task.go:134] [INFO] "apply volume, it will take about 10 seconds"
[migrate/task.go:140] [INFO] "apply volume done"
```

Then upload verification fails:

```text
Error: migrate: copy to cloud migrate: copy to cloud storage: copy prefix storage: copy prefix storage: copy prefix storage: server copier verify copy: storage: aws head object Access Denied.
storage: server copier verify copy: storage: aws head object Access Denied.
```

The migration job is never created by the CLI.

## Additional validation

Using the same Stage credentials returned by `/v2/stages/apply`:

- `PutObject` succeeds.
- `HeadObject` fails with `403 Forbidden`.

When we manually upload the same backup objects to the returned `uploadPath` and skip `HeadObject`, then call `/v2/migrations/startByRemote`, the migration job is created and completes successfully:

```text
JOB status=SUCCESSFUL progress=100
```

So the backup content and backend migration path are valid; the blocking point is the CLI-side `HeadObject` verification.

## Expected behavior

`milvus-backup migrate` should be able to complete in Stage-upload scenarios where the temporary Stage credentials are write-only, or should provide a supported way to skip client-side `HeadObject` verification and rely on the backend migration job to validate uploaded files.

Possible fixes:

1. Add a migrate option to skip upload `HeadObject` verification for Stage destinations.
2. Treat `HeadObject AccessDenied` after successful `PutObject` as non-fatal for Stage uploads.
3. Verify upload using `PutObject` response / uploaded byte count instead of requiring `HeadObject`.
4. Alternatively, document that Stage credentials must include read/head permission and fail early with a clearer error.

## Notes

This was reproduced twice with two independent backups:

- `regeneron_restore_src_20260520_backup`
- `regeneron_round2_20260520_backup`

Both backups were valid locally, and the second one completed successfully through a manual Stage upload + `startByRemote` flow.

huanghaoyuanhhy:
## Summary

Fixes #1050

Per-object `HeadObject` verification after each copy cost one API request per file and required `HeadObject`/`GetObject` permission — which is exactly why `migrate` failed against a Zilliz Cloud Stage whose credentials grant `PutObject`/`ListObject` but deny `HeadObject`.

This replaces it with a standalone `VerifyPrefixTask` that lists a destination prefix once and checks every expected key exists with a matching size, cutting the request count from O(N) to O(N/1000) (`ListObjects` returns up to 1000 keys per request) and needing only `ListObject` permission.

Copy tasks (`CopyPrefixTask`/`CopyObjectsTask`) no longer verify — they only copy. The orchestrating layer runs `VerifyPrefixTask` after the copy, where operating in terms of prefixes is natural.

## Changes

- Add `VerifyPrefixTask` (`internal/storage/verify_task.go`): recursively lists a prefix once and verifies every key in `Expected` exists with a matching size.
- Add `MirrorExpected` helper (`internal/storage/funcs.go`): lists a source prefix and maps keys to destination keys (skipping directory markers), to build the expected set for prefix copies.
- Remove per-file `HeadObject` verification and the `disableVerifyCopy`/`DisableVerifyCopy` flag from the copiers and copy tasks.
- `backup`: verify per collection over the `insert_log`/`delta_log` dirs after all segments are copied.
- `migrate`/`restore`/`check`: build the expected set from the source prefix and verify the destination prefix after copy. This restores copy verification for the cloud staging volume, which only grants `ListObject`.
- Update `copier_test.go` and add `verify_task_test.go`.

### Note

Previously a size mismatch triggered a per-file retry inside the copier. With batched verification, a mismatch or missing object now fails the whole verify step (no automatic re-copy), surfacing the problem loudly for a re-run.

### DIFF
--- a/core/backup/coll_dml_task.go
+++ b/core/backup/coll_dml_task.go
@@ -367,7 +367,59 @@ func (dmlt *collDMLTask) Execute(ctx context.Context) error {
 		return fmt.Errorf("backup: backup segments %w", err)
 	}
 
+	if err := dmlt.verifySegmentsData(ctx, describe.CollectionID, segments); err != nil {
+		return fmt.Errorf("backup: verify segments %w", err)
+	}
+
 	dmlt.taskMgr.UpdateBackupTask(dmlt.taskID, taskmgr.SetBackupCollDMLDone(dmlt.ns))
+
+	return nil
+}
+
+// verifySegmentsData verifies that every copied binlog exists in the backup
+// storage with a matching size. The destination keys all live under the
+// collection-level insert_log and delta_log dirs, so two recursive lists cover
+// the whole collection instead of issuing a HeadObject per binlog.
+func (dmlt *collDMLTask) verifySegmentsData(ctx context.Context, collID int64, segments []*backuppb.SegmentBackupInfo) error {
+	insertExpected := make(map[string]int64)
+	deltaExpected := make(map[string]int64)
+	for _, seg := range segments {
+		insertAttrs, err := dmlt.insertLogsAttrs(seg)
+		if err != nil {
+			return fmt.Errorf("backup: build insert log expected %w", err)
+		}
+		for _, attr := range insertAttrs {
+			insertExpected[attr.DestKey] = attr.Src.Length
+		}
+
+		deltaAttrs, err := dmlt.deltaLogAttrs(seg)
+		if err != nil {
+			return fmt.Errorf("backup: build delta log expected %w", err)
+		}
+		for _, attr := range deltaAttrs {
+			deltaExpected[attr.DestKey] = attr.Src.Length
+		}
+	}
+
+	insertDir := mpath.BackupInsertLogDir(dmlt.backupDir, mpath.CollectionID(collID))
+	insertVerify := storage.NewVerifyPrefixTask(storage.VerifyPrefixOpt{
+		Cli:      dmlt.backupStorage,
+		Prefix:   insertDir,
+		Expected: insertExpected,
+	})
+	if err := insertVerify.Execute(ctx); err != nil {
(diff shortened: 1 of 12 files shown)


lentitude2tk (approved):
LGTM

PR #1053, opened 2026-05-22
fix(deps): bump google.golang.org/api to v0.259.0 (fix gcpnative 'multiple credential options provided')

DashUmEr:
## Problem
The `gcpnative` storage client fails at init with:
```
storage: create gcp native client: dialing: multiple credential options provided
```
when milvus-backup is configured with explicit GCP credentials (a JSON key →
`option.WithCredentials`) in an environment that **also** has ambient Application
Default Credentials — e.g. running on **GKE with Workload Identity**. The option
validator in `google.golang.org/api` v0.258.0 treats the explicit credential
option as a conflict.

This is **independent of #1047** (the standard-GCS endpoint fix): with #1047 in
place, `WithEndpoint` is correctly skipped for the standard host so only
`option.WithCredentials` is passed — yet client init still fails on the
validation in v0.258.0.

## Fix
Bump `google.golang.org/api` v0.258.0 → v0.259.0, which includes
[googleapis/google-api-go-client#3420](https://github.com/googleapis/google-api-go-client/pull/3420)
("remove `option.WithAuthCredentials` from validation"). `go mod tidy`
regenerated; `go build ./...` passes.

## Verification
A milvus-backup build with this bump runs `gcpnative` backups to GCS
successfully in production (GKE + Workload Identity + a mounted JSON key).
Without it, the same configuration fails at client init as above.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	go.uber.org/zap v1.27.1
 	golang.org/x/oauth2 v0.34.0
 	golang.org/x/sync v0.19.0
-	google.golang.org/api v0.258.0
+	google.golang.org/api v0.259.0
 	google.golang.org/grpc v1.79.3
 	google.golang.org/protobuf v1.36.11
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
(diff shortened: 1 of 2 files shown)


DashUmEr:
C:/Program Files/Git/assign @lentitude2tk

PR #1055, opened 2026-05-26
backup: scope index extra etcd scan to backed-up collections

huanghaoyuanhhy:
## Summary
Fix a nil pointer panic (SIGSEGV) in the index extra info backup task. The task scanned the whole instance's `meta/field-index/` etcd prefix but matched the results only against the backed-up collections. A non-deleted index belonging to a collection outside the backup had no matching `IndexInfo`, so `addIndexExtraInfo` dereferenced a nil pointer and panicked.

## Changes
- `coll_index_extra_task`: scope the etcd scan per backed-up collection using the `field-index/{collectionID}/` key prefix, so indexes of collections outside this backup are never fetched.
- `metaBuilder`: add `backupCollectionIDs()`; make `addIndexExtraInfo` return an error instead of panicking or silently skipping when an index of a backed-up collection has no matching `IndexInfo` — a missing index would produce an incomplete backup that breaks CDC secondary restore.
- Add unit tests covering merge, unknown collection, missing index, and the per-collection scan scoping (out-of-scope collections are not fetched, deleted indexes skipped).

/kind improvement

### DIFF
--- a/core/backup/coll_index_extra_task.go
+++ b/core/backup/coll_index_extra_task.go
@@ -35,28 +35,38 @@ func newCollIndexExtraTask(taskID string, kv clientv3.KV, etcdRootPath string, m
 }
 
 func (ciet *collIndexExtraTask) Execute(ctx context.Context) error {
-	prefix := fmt.Sprintf("%s/meta/field-index/", ciet.etcdRootPath)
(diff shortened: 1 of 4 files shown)


lentitude2tk (approved):
LGTM

PR #1058, opened 2026-05-29
milvus: collapse four-part hotfix versions for feature detection

huanghaoyuanhhy:
## Summary
`parseVersionForFeature` parses the server's `BuildTags` with `StrictNewVersion`, which only accepts `MAJOR.MINOR.PATCH`. Four-part version strings (e.g. the real `v2.4.0.1-gpu-beta` Docker image tag, or internal/enterprise patch-line versions like `v2.3.22.6`) fail strict parsing and fall back to the `_latestDevVersion` sentinel (99.0.0). That silently enables every `>=` gated feature — e.g. `FlushAll` (>= 2.6.11) and `MultiL0InOneJob` (>= 2.6.5) — on a server that does not implement them, leading to runtime `Unimplemented` errors.

These four-part versions are an image/release-layer construct rather than source release tags (`git describe` never emits a four-part dotted version), but the backup tool should stay robust against whatever `BuildTags` the server reports.

## Changes
- collapse four-part (or longer) versions to their `MAJOR.MINOR.PATCH` release base before strict semver parsing, e.g. `v2.4.0.1-gpu-beta` -> `2.4.0`; the dropped trailing component takes any prerelease suffix with it
- leave the existing dev-build fallback untouched: tags like `2.6-20260404-31fb3fc` and `master-20260226-abcdef` have fewer than four dot-separated parts, so they still fall through to `_latestDevVersion`
- add tests covering real (`v2.4.0.1-gpu-beta`) and synthetic four-part versions

/kind improvement

### DIFF
--- a/internal/client/milvus/grpc.go
+++ b/internal/client/milvus/grpc.go
@@ -471,7 +471,12 @@ func (g *GrpcClient) checkFeature(ctx context.Context) error {
 // "2.6-20260404-31fb3fc" as "2.6.0-20260404-31fb3fc" — a prerelease of 2.6.0 — which
 // compares LESS than 2.6.x and silently disables features on dev builds.
 func (g *GrpcClient) parseVersionForFeature(ver string) *semver.Version {
-	sem, err := semver.StrictNewVersion(strings.TrimPrefix(ver, "v"))
+	v := strings.TrimPrefix(ver, "v")
+	// Collapse four-part hotfix versions to their release base, e.g. "2.3.22.6" -> "2.3.22".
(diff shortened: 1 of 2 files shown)


wenhuiZilliz (approved):
/lgtm

Clean fix for four-part hotfix version parsing. The `SplitN + Join[:3]` approach correctly collapses versions like `v2.4.0.1-gpu-beta` → `2.4.0` without affecting dev-build or standard semver paths. Good test coverage with real-world GPU-beta tags.

PR #1060, opened 2026-06-02
check: print effective config at start of check run

### LINKED ISSUE #1059: [Enhancement]: print effective config summary as part of 'check' output

When running `milvus-backup check`, also print the effective configuration (or at least a key summary: Milvus address, storage type, bucket / root path) before running the connectivity tests.

`check config` would stay as-is for the pure, network-free config dump.

### Why is this needed?

When helping users troubleshoot, the first two things we need are:
1. What configuration is actually in effect, and where each value came from (config file / env / override / default).
2. Whether milvus-backup can connect to Milvus and the object storage.

Today that's two separate commands (`check` + `check config`). Folding the config summary into `check` makes a single command give the full diagnostic picture, which means fewer round-trips when supporting users.

A common case: a user intends to write backups to a locally attached volume, but the storage section is still pointing at S3/MinIO. Surfacing the storage type and target up front in `check` makes that misconfiguration obvious immediately, instead of looking like an S3 permissions problem.

We keep `check config` as a separate subcommand so the offline "just show me the config" use case (no network calls, safe to run anywhere) is preserved.

### Anything else?

Related to #637 (Improve Backup User Experience).

huanghaoyuanhhy:
## Summary

`milvus-backup check` echoed a small, hardcoded storage summary, but only at the very end — after all connectivity tests. More importantly, both the CLI and the server construct the Milvus/storage clients *before* running the checks, and `milvus.NewGrpc` dials eagerly. So when Milvus or the object storage is unreachable (the most common failure), the command aborts during client construction and the user never sees what configuration was actually in effect — which is usually the first thing needed to diagnose it.

This change prints the full effective configuration table (the same one as `check config`: each value plus its source — override / env / config / default; secrets masked) **before any client is constructed**. So even when the connection fails, the resolved config and its sources are visible.

A common case this helps: a user intends to write backups to a locally attached volume, but the storage section still points at S3/MinIO. Surfacing the storage type and target up front makes that misconfiguration obvious immediately, instead of looking like an S3 permissions problem.

Closes #1059

## Changes

- Add `cfg.Config.WriteTable(io.Writer)` as the single source of truth for rendering the config table, and use it from `check config`.
- Print the config table in the `check` command before constructing any client (so it shows even when the connection fails).
- Lead the server `/check` response with the same table.
- Drop the now-redundant hardcoded storage summary from the `check` result output.
- Move the config-table tests into the `cfg` package and cover the command-level helper.

/kind improvement


lentitude2tk (approved):
LGTM

lentitude2tk (approved):
/LGTM

PR #1074, opened 2026-07-10
feat: add l0compact command to fold backup L0 segments into per-segment deltalogs

bigsheeper:
issue: #1075

## What

Adds a `milvus-backup l0compact` command that offline-folds a backup's **L0 (delete-only) segments** into per-data-segment deltalogs, then drops the L0 segments from the meta. The washed backup restores through the normal per-segment delta path — no L0 import needed.

```
milvus-backup l0compact --name <backup> [--output <name>_l0compacted]
```

The source backup is untouched; output goes to a new backup dir (re-runnable, revertible).

## Why

Under import replication, restore stamps a `commit_timestamp` on data (insert) segments, while L0 deletes carry their original per-record timestamp. For a backup that already contains L0 segments, this makes `delete_ts < insert_ts`, so deletes are silently dropped and rows that should be gone survive. Existing backups can't be re-compacted at the source.

`l0compact` performs the equivalent of an offline L0 compaction: it fans each L0 segment's deletes into the deltalogs of the data segments that actually contain those primary keys — the same fan-out shape as Milvus's own L0 compaction (add delta only; never rewrite insert data; never hard-delete rows). On restore, `FilterWithDelete` drops the matching rows at read time — behavior every worker version already has.

## How

- **Pure-Go codec** (`internal/l0compact/`, no cgo, no loon manifest):
  - v1 legacy event envelope (`magic 0xfffabc` + descriptor + event header) wrapping parquet payloads;
  - v2/v3 read as plain parquet, locating columns by the `PARQUET:field_id` metadata Milvus stamps on every column (backup stores no manifest — v2 and v3 are read identically);
  - deltalog read/write (v1 json + multi-field variants, v2 two-column pk+ts);
  - insert PK-column read (v1 column 0; v2/v3 scan column groups for the PK field id).
- **Exact PK-membership routing** (`core/l0compact/`): backups carry no statslog/bloom filter, so each data segment's PK column is read and deletes are routed only to segments that contain the key — no broadcast. Partition-level L0 folds into its partition; channel-level L0 folds into data segments of the same vchannel.
- **Meta**: `internal/meta.Write` persists the new backup (`full_meta.json` + the four leveled files); L0 segments are dropped so restore doesn't re-issue an L0 import.

Milvus itself is not modified.

## Testing

- Unit round-trip tests for every codec piece: v1/v2 deltalog read/write (Int64 and VarChar keys, json and multi-field variants), and v1/v2/v3 insert PK-column reads (incl. `PARQUET:field_id` column location).
- Integration test builds a synthetic backup, runs `Execute`, and asserts the L0 delete is folded into the correct data segment's new deltalog and the L0 segments are dropped from the meta.
- `go build ./...`, `go vet`, `gofmt` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


czs007:
## Adversarial Review Summary

The L0-compact path is functionally sound but has one correctness gap in the v1 deltalog reader plus several robustness, validation, and safety issues around backup output handling; details grouped by severity below.

### 🔴 High

- **`internal/l0compact/deltalog.go:35-46` — v1 deltalog reader only parses the first delete event.** The `KindV1` branch takes `payload := events[0].Payload` and discards `events[1:]`, even though `parseV1Envelope` (`envelope.go:90-113`) decodes all data events and the sibling `ReadInsertPK` (`insertpk.go:12-26`) iterates over all of them. A v1 L0 deltalog carrying multiple delete events would fold only the first; after `dropL0` removes the original L0 segment, the remaining deletes are lost permanently and deleted rows resurrect on restore. Iterate over all events and accumulate. *(raised by yezh, wzy)*

### 🟠 Medium

- **`cmd/l0compact/l0compact.go:22-30` — `validate()` allows `output == name` and never checks the destination, silently corrupting backups.** It only fills a default when `output` is empty. `-n bk -o bk` makes `srcDir == dstDir`, so `copyBinlogs` self-copies and `meta.Write` overwrites the source meta, breaking the "source backup is untouched" guarantee; pointing `-o` at another existing backup name unconditionally overwrites its five meta files. Reject same-name and require an explicit `--force` when the destination already exists. *(raised by yezh, wzy)*

- **`core/l0compact/task.go:267-281` — v2 PK read loads every column group into memory.** The v1 branch (`259-266`) selects only the PK-field binlog, but the v2 branch collects *all* fields/column groups and reads each parquet blob fully into `blobs` before `ReadInsertPK` scans for the PK column. For a large segment this holds the entire segment's parquet (including vector columns) in memory at once. Restrict v2 reads to the PK column group and stop once the PK is found. *(raised by wzy)*

- **`internal/l0compact/parquet.go:143,148,160,194 — unchecked type assertions can panic on malformed input.** The read path performs bare assertions (`chunk.(*array.Int64)`, `chunk.(*array.String)`, etc.) on blobs sourced from external Milvus binlogs. A corrupted file or schema drift where the stored column type differs from the expected `PKType`/schema panics instead of returning an error. Use the comma-ok form and return an error on mismatch. *(raised by wzy)*

- **`tests/testcases/test_l0compact.py:406-449` — the e2e suite cannot prove the fix addresses the missed-delete scenario.** Cases 1–8 (`_wash_restore_assert`, `:175-194`) only assert that a washed backup restores with deletes applied; they would pass even if l0compact were a no-op. The PR's motivating scenario (import replication overwriting `commit_timestamp` so `delete_ts < insert_ts` and deletes are silently dropped) is never constructed, and the only unwashed-contrast case (`test_l0compact_unwashed_contrast`) uses try/except with `log.info` and zero assertions. Add a case that reproduces `delete_ts < insert_ts` (or L0 import rejection) and asserts unwashed restore is wrong/rejected while washed restore is correct. *(raised by wzy)*

### 🟡 Low

- **`core/l0compact/task.go:38,56` — output backup meta still declares the source backup's Name/Id.** `Execute` reads the source meta, never updates `Name`/`Id`, and writes it verbatim to `dstDir`, so `bk_l0c` reports itself as `bk` with the same Id. `list` then shows two identically named/id'd backups, risking mis-restore or mis-deletion of the original. Set `info.Name` to `o.output` (and assign a new Id or define explicit retention semantics) before writing, and add a list/meta assertion test. *(raised by yezh)*

- **`core/l0compact/task.go:32,172` — `nextID` starts at 1 and can collide with an existing deltalog's logID.** New deltalogs are written with monotonically increasing IDs from 1, sharing the `deltaKey` namespace with source deltalogs copied under their original logIDs. If an existing data-segment deltalog happens to use logID 1/2/3…, the new write overwrites it. Real Milvus logIDs come from a global allocator (large int64), so collision is unlikely in practice, but the code offers no guarantee. Start from `max(existing logID) + 1`. *(raised by yezh, wzy)*

- **`core/l0compact/task.go:47-58` — a mid-fold failure leaves a meta-less half-backup in the destination.** `Execute` runs `copyBinlogs` first and `meta.Write` last; if any `foldIntoSegment` errors before the meta write, the destination holds a full `binlogs/` tree plus partially folded deltalogs but no meta — garbage requiring manual cleanup, and a re-run with the same `-o` stacks on top. Clean up the destination on failure, or verify it is empty before writing. *(raised by wzy)*

- **`core/l0compact/task.go:66-78` — the washed output is a full second physical copy of the source insert data.** `copyBinlogs` server-side-copies the entire `binlogs/` subtree, including all insert logs, so for a TB-scale backup the l0compact output roughly doubles object-storage usage. This is a reasonable "new backup, revertible" trade-off, but the command help (`l0compact.go:52,62`) does not mention the space cost. Document it. *(raised by wzy)*

bigsheeper:
Thanks for the thorough adversarial review — great catches. Addressed in `cfa7f95` (fixes) + `b78b08c` (regression test):

### 🔴 High
- **v1 deltalog reader only parsed the first event** — Fixed. `ReadDeltalog` KindV1 now iterates **all** parsed events and accumulates (json or MULTI_FIELD per event). Added `TestDeltalogV1MultiEvent`, a two-event fixture verified to fail against the old single-event read and pass after.

### 🟠 Medium
- **`validate()` allowed `output == name` / silent overwrite** — Fixed. Rejects `output == name`; added `--force`; `run()` errors on an existing destination unless `--force`.
- **v2 PK read loaded every column group into memory** — Fixed. v2 now probes one group file at a time (`ReadParquetColumnByFieldID`) and stops at the group carrying the PK column; peak memory is a single group file. v1 unchanged.
- **Unchecked type assertions can panic** — Fixed. The parquet read path uses comma-ok assertions and returns a descriptive error on type mismatch.
- **e2e cannot prove the missed-delete scenario** — Acknowledged, in progress. This is tied to a restore integration issue the L2 e2e run surfaced (a folded deltalog is read with the wrong version on restore); I'm root-causing that first, then will add a negative case that asserts the unwashed restore is wrong/rejected while the washed restore is correct.

### 🟡 Low
- **Output meta kept the source Name/Id** — Fixed. Sets `info.Name = <output>` and a fresh `uuid` Id before writing.
- **`nextID` could collide with an existing logID** — Fixed. Starts from `max(existing logID) + 1` across all binlogs/deltalogs.
- **Mid-fold failure left a meta-less half-backup** — Fixed. `Execute` requires an empty destination and best-effort cleans it up on any failure.
- **Full second physical copy / space cost undocumented** — Fixed. Command help now states the output is a complete physical copy (~doubles object-storage usage).

czs007:
## Adversarial Review

Adversarial review surfaced three issues in the new `l0compact` command — two `medium` correctness/security defects in argument handling and one `low` test-coverage gap that can silently mask regressions.

### Medium

**1. `cmd/l0compact/l0compact.go:24` — `--name` / `--output` are not validated, allowing path traversal outside the backup root**

*Problem:* `options.validate()` (lines 24–35) only checks that `name` is non-empty and that `output != name`. Unlike `create` (`cmd/create/create.go:61`, `core/server/create.go:224`), it never calls `backup.ValidateName`, which rejects blanks, special characters, and names not starting with a letter/underscore (so `/`, `.`, `..` are all refused). Since `mpath.BackupDir` is `path.Join(backupRoot, name) + "/"` and `path.Join` cleans the path (`internal/storage/mpath/path.go:106-108`), `--output ../victim` resolves to a prefix *outside* the backup root (another prefix in the same bucket), `--output a/b` nests the new backup underneath an existing one, and `--name ../x` lets the read path escape the root the same way. The resulting directory cannot be managed as a normal backup by `list` / `get` / `delete` / `restore`.

*Secondary risk:* on failure, `core/l0compact/task.go:68` runs `DeletePrefix(t.dstDir)` against that out-of-root prefix. The "destination must be empty" guard at `task.go:48-54` means this cannot delete pre-existing objects, so it is a secondary concern rather than the main one — the primary problem is the out-of-root read/write and the unmanageable backup it produces.

*Suggestion:* call `backup.ValidateName` on both `o.name` and the defaulted `o.output` inside `validate()`, and add unit tests for `--output ../x` and `--output a/b`.

*(raised by yezh)*

**2. `cmd/l0compact/l0compact.go:45` — `--force` can never overwrite an existing backup**

*Problem:* The `meta.Exist` rejection is gated behind `if !o.force` (lines 45–53), but the destination-not-empty check in `core/l0compact/task.go:48-54` is unconditional: `Execute` starts with `ListPrefixFlat(dstDir)` and returns an error whenever `len(keys) > 0`. `Task` has no `force` field at all (`NewTask(cli, srcDir, dstDir)`, `task.go:35`), and `force` is never threaded through (`l0compact.go:54`). Any genuinely existing backup has at least `meta/full_meta.json` and `binlogs/*` under `dstDir`, so `--force` merely skips the cmd-layer check and then unavoidably trips the task-layer check. The help text claims it will "overwrite the destination backup if it already exists" (`l0compact.go:81`), which never holds — and leftover half-written directories from a previous failed run cannot be cleaned up by re-running with `--force` either.

*Suggestion:* either thread `force` into `Task` and `DeletePrefix(dstDir)` before the emptiness check in `Execute`, or drop the flag and its help text. Add a regression test covering the "destination already exists" path.

*(raised by wzy, yezh)*

### Low

**3. `tests/testcases/test_l0compact.py:204` — the only test that directly asserts L0 removal from the meta can silently skip**

*Problem:* `_read_full_meta` (lines 203–229) wraps every MinIO error in `except Exception as e: pytest.skip(...)`, and its sole consumer is `test_l0compact_structural_meta` (case 8, lines 391–404, calls at 399/400). That case is the only assertion in the suite that directly verifies the `dropL0` result landed in the meta JSON (`_meta_has_l0` checks `collection_backups[].l0_segments` and `is_l0` segments inside partitions). The endpoint, bucket, and credentials are hardcoded to `localhost:9000` / `a-bucket` / `minioadmin`. On any CI environment that doesn't match exactly, case 8 skips silently and is counted as a pass — so a regression that leaves `l0_segments` / `is_l0` behind would go unreported, relying only on the indirect restore-based cases.

*Suggestion:* fail rather than skip when MinIO is unreachable (or restrict the skip strictly to the "object genuinely does not exist" case), and move the endpoint/bucket/credentials into the test configuration.

*(raised by wzy)*

bigsheeper:
Thanks — all three addressed in `d134f89`:

**Medium 1 (name/output not validated → path traversal)** — `validate()` now calls `backup.ValidateName` on both `--name` and the (defaulted) `--output`, same as `create`, so `../x`, `a/b`, etc. are rejected. Added `TestValidateRejectsBadNames`.

**Medium 2 (`--force` could never overwrite)** — Fixed the dead flag: `force` is now threaded into `Task` (`WithForce`), and `Execute` clears a non-empty destination (`DeletePrefix`) when force is set instead of unconditionally erroring; the redundant cmd-layer `meta.Exist` check was removed so destination handling lives in one place. Added `TestExecuteForceClearsDestination` covering both the reject (no force) and clear-then-succeed (force) paths.

**Low 3 (structural test silently skips)** — `_read_full_meta` now only `pytest.skip`s on a genuine `NoSuchKey`/`NoSuchBucket`; any other MinIO error (e.g. unreachable) raises and fails the test, so a `dropL0` regression can no longer hide behind a silent skip.

czs007:
Adversarial review surfaced two low-severity issues in the new l0compact path — both are latent rather than actively breaking, but worth tightening before merge.

### Low

**`internal/l0compact/deltalog.go:105` — `KindV1` write path discards `json.Marshal` errors**

The `KindV1` branch of `WriteDeltalog` (lines 96–107) drops three `json.Marshal` errors: `pkRaw, _ = json.Marshal(e.PK.Int)` (:101), `pkRaw, _ = json.Marshal(e.PK.Str)` (:103), and `b, _ := json.Marshal(deleteLogJSON{...})` (:105). With the current payload types (`int64` / `string` / `uint64`) marshalling cannot fail, so this is harmless today. But `deleteLogJSON.Pk` is a `json.RawMessage`, so if the PK types or fields are extended later, a marshal failure would be silently swallowed and emit `rows[i] = ""` — producing a deltalog that writes successfully but cannot be read back.

*Suggestion:* propagate the error explicitly (`return err`), or at minimum add a comment documenting why it is safe to ignore given the closed set of PK types. (Note: the originally reported line 1024 does not exist — the file is 125 lines.)

*(raised by wzy)*

**`core/l0compact/task.go:262` — new fold deltalog stores the backup-internal key in `Binlog.LogPath`, contradicting the repo-wide convention**

`foldIntoSegment` records the newly written deltalog as `&backuppb.Binlog{LogId: logID, LogPath: key, ...}` (lines 256–263), where `key = deltaKey(t.dstDir, ...)` is an object key inside the backup. The comment in the same file at lines 314–317 states the opposite convention: "The stored `Binlog.LogPath` is the ORIGINAL milvus source path, not the backup's own key, so reads must reconstruct". The backup side follows that convention — `core/backup/coll_dml_task.go:473` does `srcAttr := storage.ObjectAttr{Key: binlog.LogPath}`, treating `LogPath` as a Milvus source object. As a result, a single `Deltalogs` array in the compacted meta ends up mixing two semantics: copied entries hold Milvus source paths, while newly folded entries hold backup keys.

Nothing breaks today (grepping the repo, `LogPath` is only consumed by the `core/backup` copy path; neither restore nor meta reads it), but it is a semantic trap for any future code that reads or validates by `LogPath`.

*Suggestion:* either leave `LogPath` empty or populate it with the same "original source path" semantics, and annotate the field at the assignment site.

*(raised by zack)*

bigsheeper:
Both addressed in `caa33be`:

- **`WriteDeltalog` KindV1 discarded `json.Marshal` errors** — now propagated (`return err` on pk marshal and DeleteLog marshal), so a future PK-type extension can't silently emit an empty, unreadable row.
- **Fold deltalog `LogPath` semantics** — annotated at the assignment site: copied entries carry the original Milvus source path, while a newly folded deltalog only exists inside the backup, so its `LogPath` is the backup object key; restore reconstructs delta paths from the backup dir and never reads `LogPath`, so it's inert for restore.

Separately, the same commit fixes the real reason the L2 e2e was failing (surfaced by CI, not this review): Milvus's import reads a segment's deltalogs with the **v1** binlog reader even on the v2 restore path, so the fold deltalog must be written in the v1 envelope format regardless of segment storage version (a bare v2 parquet deltalog fails "parse magic number"). Insert data is still read per the segment version.

bigsheeper:
### CI status (commit `caa33be`)

The l0compact e2e suite (tag **L2**) now **passes on both `2.6-latest` and `master-latest`**, so the end-to-end path is validated on real Milvus: build a backup containing L0 → `l0compact` → restore → deleted rows are correctly absent. On `2.6-latest` the whole `api` matrix (L0/L1/L2/RELEASE/RBAC) is green.

The remaining red checks are **not related to this PR**:

- **`master-latest` L1 / MASTER / RELEASE** — pre-existing failures in `tests/testcases/test_restore_backup.py` (`test_milvus_restore_back_with_delete`, `_with_dup_pk`, `_with_upsert`), i.e. general backup/restore cases that assert restored data consistency. They fail only on `master-latest` (the same cases pass on `2.6-latest`) and touch none of the l0compact code — this is Milvus-master instability, not this PR. l0compact's own tests are tagged L2, which is green on master.
- **`secondary (same-version)`** — a flaky `schema version is not consistent` timeout restoring the example `hello_milvus` dataset; it has failed on every run of this branch and is unrelated to l0compact.

czs007:
## Overview
Adversarial review surfaced five issues in the L0 compaction path — one medium-severity memory-footprint concern in the v2 insert PK read path, plus four low-severity items covering backup-layout compatibility, a silent-failure asymmetry, orphaned objects, and redundant per-partition work; none are blocking, but the medium item is worth converging before merge.

---

### Medium

**`internal/l0compact/parquet.go:129` — v2 insert PK read path buffers the whole object and materializes the full table**

*Problem.* `readSegmentPKs` (`core/l0compact/task.go:367-391`) calls `storage.Read` to pull the **entire column-group object** into memory, then `ReadParquetColumnByFieldID` → `readTable` → `r.ReadTable(ctx)` materializes **every column** of that group — including large vector columns — into an `arrow.Table`. Peak memory is therefore roughly (raw group-file bytes) + (fully decoded Arrow table), not the "peak memory to a single group file" claimed in the comment at `core/l0compact/task.go:345-349`. If the real v2 layout co-locates the PK column and vector columns in the same column group, a single group can exhaust memory when compacting a large backup offline.

*Suggestion.* Arrow-side column projection alone is insufficient — the whole-object buffer in `storage.Read` remains. To actually bound the footprint, use a ranged/streaming read plus PK-only projection (locate the column index via `PARQUET:field_id`, then `r.GetColumn` / a row-group reader), or consume batches through a record reader.

*(raised by yezh, wzy)*

---

### Low

**`core/l0compact/task.go:327` — key reconstruction always appends GroupID, missing the group-less layout that restore still supports**

*Problem.* `insertKey()` unconditionally passes `mpath.GroupID(seg.GetGroupId())`, and `deltaKey()` appends GroupID for every non-`allPartitionID` segment, so both only match backups stored as `coll/part/group/seg` with a non-zero GroupId. Restore, however, still keeps an explicit group-less compatibility branch (`notL0SegBatchesWithoutGroupID`, `core/restore/coll_task.go:540`, entered from `notL0SegmentBatches` when all non-L0 segments have `GroupId == 0`, `coll_task.go:648-661`). Since `mpath` writes a literal `"0"` path element for `GroupID(0)` (`internal/storage/mpath/path.go:138-150`), l0compact would build `binlogs/insert_log/<coll>/<part>/0/<seg>/...` while the object actually lives at `.../<coll>/<part>/<seg>/...`, producing a hard key-not-found on PK read rather than a silent miscomputation.

The trigger surface is narrow: the group-less layout only comes from pre-2023-10 writers (old `copySegment` inserted the group directory only `if segment.GetGroupId() != 0`), while L0 backup support only landed in 2024-06 (`aaadb04`), so the two rarely coexist in one backup. This is an alignment/hardening item, not a reproducible production failure.

*Suggestion.* Either make key reconstruction layout-aware (fall back to the group-less path when `GroupId == 0`), or explicitly reject group-less backups at the `Execute` entry point instead of letting them fail as an opaque not-found.

*(raised by yezh)*

---

**`internal/l0compact/insertpk.go:12` — v1 path silently returns empty when the PK binlog is missing, unlike v2**

*Problem.* The v1 branch of `readSegmentPKs` only collects binlogs whose `FieldID == pkFieldID`. If a segment's meta carries no `FieldBinlog` for the PK field, `blobs` is empty and the `KindV1` branch of `ReadInsertPK` performs no `found` check — it returns `nil, nil`. The `KindV2` branch, by contrast, errors explicitly when no group contains the PK column. The asymmetry means such a segment folds to zero hits, `foldIntoSegment` returns early without writing a deltalog (`core/l0compact/task.go:236-253`), yet `run()` still calls `dropL0` unconditionally (`task.go:99-102`) — yielding a backup that has **lost its delete semantics** with no error surfaced. Real segments always carry a PK binlog, so the trigger surface is narrow, but the failure mode is a silent correctness bug rather than a hard failure.

*Suggestion.* Make the v1 branch match v2: return an error when the PK field is not found.

*(raised by yezh, wzy)*

---

**`core/l0compact/task.go:152` — `copyBinlogs` copies the L0 delta subtree, leaving orphaned objects**

*Problem.* `copyBinlogs` runs `CopyPrefix` over the entire `binlogs/` prefix, so L0 segments' `delta_log` subtrees are copied into the destination backup as well. `dropL0` then removes those segments from the meta (`core/l0compact/dropl0.go:13-21`) without deleting the objects, leaving a set of objects in the destination that no meta references. The PR's own test acknowledges this (`task_test.go:47-49`: "The original L0 deltalog is also copied under dstDir (harmless after drop)"). l0compact is already a full physical copy — the CLI's long help states it doubles storage — so this is incremental waste on top of that, not a new class of cost.

*Suggestion.* Exclude L0 segments' delta prefixes during the copy based on the meta, or clean them up after the fold completes.

*(raised by wzy)*

---

**`core/l0compact/task.go:180` — channel-level L0 deltalogs are downloaded and parsed once per partition**

*Problem.* `foldCollection` calls `buildDeleteMap` inside the partition loop, and the `l0segs` it passes include the collection-level (channel-level) L0 segments returned by `channelL0For(coll, targets)` (`task.go:308-320`). The same channel L0 deltalog objects are therefore re-read via `storage.Read` and re-decoded via `ReadDeltalog` once per partition, and the identical `pk -> ts` map is rebuilt each time (`task.go:196-215` — real object reads and decoding, no caching). For a collection with P partitions sharing one vchannel, download and decode cost for channel L0 is amplified P×, with the duplicated `delMap` also held in memory. Behavior is correct; this is pure overhead.

*Suggestion.* Build the channel-level delete map once at collection scope (cached per vchannel) and only overlay partition-level L0 inside the partition loop.

*(raised by zack)*

bigsheeper:
All five addressed in `cecc2ba`:

**Medium — v2 insert PK read decoded every column** — `ReadParquetColumnByFieldID` now resolves the column index from the schema (no data read) and reads only that one column via `ReadRowGroups(ctx, []int{idx}, allRowGroups)`, so vector columns in the group are never decoded. Comment updated to be accurate: only the PK column is decoded; `storage.Read` still buffers the whole group object (a fully ranged/streaming read is noted as a follow-up).

**Low — group-less layout** — `insertKey`/`deltaKey` now append the GroupID element only when `GroupId != 0`, matching backup's own path writer and restore's group-less compatibility branch. (Test's synthetic v2 data segment now carries `GroupId == SegmentId` to mirror real v2 layout.)

**Low — v1 silent empty on missing PK binlog** — the v1 read path now errors (`PK field %d has no insert binlog in segment %d`) instead of returning empty, matching the v2 branch — so `dropL0` can't run on a segment whose deletes were silently lost.

**Low — orphaned L0 delta objects** — after `dropL0`, the dropped L0 segments' destination `delta_log` prefixes are deleted (collected before `dropL0` mutates the meta). Best-effort: a delete failure warns rather than failing the run.

**Low — channel-level L0 re-read per partition** — channel (all-partition) L0 delete maps are now built once per vchannel and cached; the partition loop only decodes partition-level L0 and overlays the cached channel map (max-ts merge), so channel L0 objects aren't re-downloaded/re-decoded P× times.

czs007:
Adversarial review found the following issues requiring changes.

## Overview
The L0 compaction backup/restore path has three robustness gaps: an unconditional L0 drop that can silently discard unfolded deletes, an unvalidated length field that can panic or over-allocate on malformed input, and a missing test for the VarChar v1-json round-trip.

## Low

**`internal/l0compact/envelope.go:78`** — `parseV1Envelope` trusts the descriptor's `extraLen` before validation. `extraLen` is an `int32` read straight from the v1 descriptor and passed to `make([]byte, extraLen)` with no bounds check: a negative value panics (`makeslice: len out of range`) and a huge positive value over-allocates (up to ~2GB) before `readFull` discovers truncation. This runs inside `Task.Execute` (`buildDeleteMap → ReadDeltalog → parseV1Envelope`) with no `recover()`, so a panic bypasses `run()`'s error-path cleanup and leaves a partial destination backup in `dstDir`. The sibling event-payload path already guards this (`envelope.go:101` rejects `payloadLen < 0`), making `extraLen` an inconsistent omission. Suggestion: reject `extraLen < 0` and cap it against `r.Len()` before allocating, returning a contextual error. (raised by yezh)

**`core/l0compact/task.go:214`** — L0 is folded only into vchannel-matching data segments, but `dropL0` strips all L0 unconditionally. `foldCollection` folds channel-level L0 into a partition only for the vchannels its data segments span (`channelMapFor` at `task.go:196` selects `coll.L0Segments` by exact `GetVChannel()` string equality), and the skip guard at `task.go:214` never asserts that every collection-level L0 segment matched at least one target. `dropL0` (`dropl0.go:15-18`) then removes all L0 unconditionally (`L0Segments = nil` plus an `IsL0` filter). So any collection-level L0 whose vchannel matches no data segment's vchannel — e.g. vchannel-string asymmetry, or a data segment with an empty `VChannel` while the L0 carries a populated one — folds into nothing yet is still dropped, and its deletes silently vanish on restore. The asymmetry precondition is unverified (fresh 2.6 e2e backups have symmetric populated vchannels, so this path is untested), so this is a defensive gap rather than a demonstrated regression. Suggestion: `log.Warn` or hard-error when a collection-level L0 segment folds into zero target data segments before `dropL0` removes it. (raised by wzy)

## Nitpick

**`internal/l0compact/deltalog_test.go:9`** — The v1-json deltalog write+read path is only unit-tested for Int64 PKs, not VarChar. The fold always writes the v1-json envelope (`foldIntoSegment` at `task.go:311` calls `WriteDeltalog` with `KindV1`) and VarChar PKs are fully supported on that path (`WriteDeltalog` marshals `e.PK.Str` at `deltalog.go:104`; `readV1JSONDeltalog` unmarshals into `pk.Str` at `deltalog.go:76`). But `TestDeltalogV1JSONRoundTrip` covers only `PKInt64`, and VarChar round-trips are exercised only for `KindV2` (`TestDeltalogV2RoundTripVarChar`). So a VarChar-PK backup hits the v1-json VarChar write/read exclusively at e2e time with no unit coverage. Suggestion: add a `KindV1` + `PKVarChar` round-trip assertion. (raised by wzy)

bigsheeper:
Thanks for the review. All three are fair hardening points; none is a demonstrated regression on real backups, so keeping this PR as-is and tracking them as follow-ups:

**`envelope.go:78` — unvalidated `extraLen`** — Agreed it's an inconsistent omission vs the `payloadLen < 0` guard at `envelope.go:101`. Triggering it requires a corrupted v1 descriptor in the source backup — every deltalog read here was written by Milvus itself (or by this tool). Worst case is a panic leaving a partial destination dir; the source backup is never touched, and the destination is a fresh dir that can simply be deleted and re-run. Will add the `extraLen < 0` / cap-against-remaining-bytes check as a follow-up hardening.

**`task.go:214` — L0 dropped even when it folded into zero targets** — Agreed as a defensive gap. For backups that contain L0 at all (L0 backup support is 2.6+), data segments and channel L0 carry symmetric, populated vchannels, so a zero-match fold hasn't been observed in practice. That said, silent delete loss is exactly what this command exists to prevent, so a follow-up will add a hard error when a collection-level L0 segment matches no target data segment before `dropL0` removes it.

**VarChar `KindV1` unit round-trip** — The path is exercised end-to-end today: e2e case 2 (`test_l0compact_varchar_pk`) folds a VarChar-PK backup (the fold always writes the v1-json envelope) and asserts deleted rows stay gone after restore. Agreed a unit-level `KindV1 + PKVarChar` round-trip is still worth having for faster signal; will include it in the same follow-up.

PR #1077, opened 2026-07-13
restore: support secondary import two-phase commit

huanghaoyuanhhy:
## Summary
Add state-based compatibility for Milvus Import 2PC during Secondary restore.

Milvus master rejects `auto_commit=true` imports in a replicating cluster, so the
import must be committed explicitly. The commit is driven by a CommitImport V2 WAL
message broadcast over the collection's data vchannels, keeping the commit ordered
inside the replicate stream rather than stamped by the target cluster's own clock.

## Changes
- disable automatic import commit for replicated Secondary import messages
- detect Uncommitted and Committing states without Milvus version checks
- broadcast wire-compatible CommitImport V2 messages over collection data vchannels
- preserve non-L0 and L0 phase barriers and per-pchannel timestamp ordering
- cover legacy and 2PC state paths, wire properties, header/body encoding, and fanout
- add a `same-version-master` scenario to the secondary restore CI job, alongside
  the existing 2.6 one, so the 2PC path is actually exercised — import 2PC only
  exists on the master line, which #1080 had dropped from the matrix

/kind improvement

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -594,8 +594,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        # Two scenarios:
+        # Three scenarios:
         #   - same-version:          source and target are both the latest 2.6.
+        #   - same-version-master:   source and target are both the latest master,
+        #     the only line that has import two-phase commit. #1080 dropped master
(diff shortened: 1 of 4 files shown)


lentitude2tk (approved):
/LGTM

PR #1079, opened 2026-07-13
restore: drop max_field_id property when creating collection

huanghaoyuanhhy:
## Summary

Restoring a collection whose fields were added via `add_collection_field` silently loses all data in those fields — they come back `NULL` for every row, with no error. Original fields round-trip correctly. This reproduces only against Milvus `master` images, which is why `master-latest` CI is red while `2.6-latest` passes.

**Root cause.** Milvus master records a server-managed `max_field_id` collection property so that field IDs are not reused after a field is dropped (milvus-io/milvus#48988). Restore replayed it verbatim into `CreateCollection` — in both `schema.Properties` and the request `Properties`, and Milvus reads it from both — and `updateMaxFieldIDProperty` keeps whichever value is larger.

For a source collection with fields `100-104`, `$meta` `105`, and added fields `106-110`, the backup carries `max_field_id=110`. On restore:

1. `CreateCollection` makes fields `100-104` + `$meta` `105`, whose real maximum is `105`, but persists the inherited `max_field_id=110`.
2. The replayed `AddField` calls compute `maxAssignedFieldIDFromSchema(schema) + 1` = `max(105, 110) + 1` → they allocate **`111-115`** instead of `106-110`.
3. The backed-up binlogs live in field-ID directories `106-110`. The import skips them because they are not in the target schema, and NULL-fills `111-115`, which are nullable and have no data. No error is raised.

Confirmed from CI logs, which show both ID sets on the same run:

```
fieldID:106..110  added_int64 … added_json   <- source collection
fieldID:111..115  added_int64 … added_json   <- restored collection
```

Milvus 2.6 has no `max_field_id` property, so `AddField` computes `105 + 1` = `106` there and the IDs line up.

## Changes

- Strip `max_field_id` from the properties passed to `CreateCollection`, both in the schema and in the request properties, and let Milvus recompute it from the restored schema
- Add unit tests covering both call sites, and assert the caller-configured `SkipParams.CollectionProperties` slice is not mutated

The secondary restore path is unaffected: it sends the full schema with explicit field IDs, so the property stays consistent with the schema there.

/kind improvement

### DIFF
--- a/core/restore/coll_ddl_task.go
+++ b/core/restore/coll_ddl_task.go
@@ -21,6 +21,19 @@ import (
 	"github.com/zilliztech/milvus-backup/internal/validate"
 )
 
+// maxFieldIDKey is a server-managed collection property that records the highest
+// field ID ever assigned, so that IDs are not reused after a field is dropped.
+// Milvus derives it from the schema when absent, and CreateCollection keeps
+// whichever value is larger, so replaying a backed-up value makes AddField
+// allocate IDs above the source's maximum instead of reproducing the source IDs.
+// The binlog directories are named after the source field IDs, so the import
+// then finds no data for the added fields and silently fills them with NULL.
+// Always drop it and let Milvus recompute it for the restored collection.
+//
+// Not defined in pkg/v2/common at the pinned Milvus version: the property only
+// exists on Milvus master (milvus-io/milvus#48988).
+const maxFieldIDKey = "max_field_id"
+
 type collDDLTask struct {
 	option       *Option
 	collOverride CollOverride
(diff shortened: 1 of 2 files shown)


huanghaoyuanhhy:
Closing for now: this only reproduces master→master, and we are dropping Milvus master from CI (#1080) and tracking master compatibility on a separate version line instead. Landing it on main would leave it with no integration coverage.

The bug is real and should be re-landed on that line — recording the analysis here so it is not lost.

**Root cause.** Milvus master records a server-managed `max_field_id` collection property (milvus-io/milvus#48988, master-only — `git grep max_field_id upstream/2.6` is empty). Restore replayed it verbatim into `CreateCollection`, in both `schema.Properties` and the request `Properties`, and Milvus reads it from both. `updateMaxFieldIDProperty` keeps whichever value is larger.

For a source with fields `100-104`, `$meta` `105` and added fields `106-110`, the backup carries `max_field_id=110`. On restore:

1. `CreateCollection` makes `100-104` + `$meta` `105` (real maximum `105`) but persists the inherited `max_field_id=110`
2. The replayed `AddField` calls compute `maxAssignedFieldIDFromSchema(schema) + 1` = `max(105, 110) + 1` → they allocate **`111-115`** instead of `106-110`
3. Backed-up binlogs live in field-ID directories `106-110`. The import skips them because they are absent from the target schema, and NULL-fills `111-115`, which are nullable and have no data. **No error is raised.**

Confirmed from CI logs, which show both ID sets in the same run:

```
fieldID:106..110  added_int64 … added_json   <- source collection
fieldID:111..115  added_int64 … added_json   <- restored collection
```

Only master→master is affected. 2.6 assigns field IDs via `nextFieldID()` (`internal/rootcoord/util.go:578`), which walks the actual fields and never reads any property, so a master-taken backup restored into 2.6 still lands on `106`.

**Impact.** Silent data loss: no error, row counts match, schema matches — only the added fields come back all-NULL. Once 3.0 ships, any `3.0 → 3.0` restore of a collection that used `add_collection_field` loses that data.

**Fix.** Strip `max_field_id` from the properties passed to `CreateCollection` (both sites) and let Milvus recompute it from the restored schema. The secondary restore path is unaffected: it sends the full schema with explicit field IDs.

Branch `strip-max-field-id` is kept, commit `bf7c6f0`.

PR #1080, opened 2026-07-13
ci: stop testing against milvus master images

huanghaoyuanhhy:
## Summary

Milvus master has diverged from 2.6 in ways that break backup and restore. The `master-latest` jobs have been red for weeks, which hides regressions in the jobs that actually gate the released line — every `2.6-latest` job passes today. Drop master from the PR/push and nightly matrices so CI reflects what we ship, and track master compatibility on a separate version line instead.

Known master-only breakage, for the record:

- Milvus master records a server-managed `max_field_id` property (milvus-io/milvus#48988), which made restore allocate the wrong field IDs and silently lose all data in fields added via `add_collection_field` (fixed separately in #1079)
- Milvus master stamps imported segments with `commit_timestamp` (milvus-io/milvus#48472), which makes segcore discard every restored delete, so deleted and pre-upsert rows reappear after restore — still open
- The `bumpSchemaVersion` compaction that replaced the always-on backfill policy defaults to disabled (milvus-io/milvus#48808), so `schema_version_consistent_segments` never converges (worked around in #1078)

## Changes

`main.yaml`
- `api`, `cross-version`, `after-upgrade`: drop the `master-latest` image
- `secondary` same-version: run on `2.6-latest` instead of `master-latest`
- `cli` and `rbac-config`: repoint to `2.6-latest`. `master-latest` was their only image, so dropping it would have removed the jobs entirely rather than just narrowing them
- `api`: drop the `MASTER` case tag, which no longer has an image to run on
- `after-upgrade`: `2.5-latest` was listed twice in `source_image_tag`, which paired with the two targets; with one target left it would have run the identical job twice

`nightly.yaml`
- drop the `master` milvus_version and the `MASTER` case tag

Both files: remove the now-dead `master-latest` branches that set `enablev2` / `useLoonFFI`.

## Coverage given up

The three MinHash cases (`test_milvus_restore_back_with_minhash_function*`) keep their `MASTER` tag but no longer run anywhere, since the feature only exists on master. They come back with the master-compatible version line.

`perf.yaml` still runs `master-latest` and is left alone — it is green and does not gate PRs. `azure.yaml` runs `2.5-latest` and is unrelated.

/kind improvement

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -77,7 +77,7 @@ jobs:
         milvus_mode: [standalone]
         another_milvus_mode: [standalone]
         source_image_tag: [v2.2.16, v2.3.22, v2.4.23, v2.5.20, 2.6-latest]
-        target_image_tag: [master-latest, 2.6-latest]
+        target_image_tag: [2.6-latest]
         exclude:
           - source_image_tag: 2.6-latest
             target_image_tag: 2.6-latest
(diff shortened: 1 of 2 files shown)


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

PR #1083, opened 2026-07-15
refactor(config): clarify storage and transfer settings

### LINKED ISSUE #1005: [Enhancement]: Misleading configuration `minio.storageType`

I found the configuration naming is confusing and misleading. For example, the configuration field minio.storageType appears to be a MinIO-specific configuration. However, in reality, it represents the storage type used by Milvus.

If minio.storageType is 'local', it has nothing to do with MinIO, but it uses the configuration under **minio** namespace.

Would it be more appropriate to align it with Milvus’s own configuration style, such as [common.storageType](https://milvus.io/docs/configure_common.md#commonstorageType), for better consistency and clarity?

In addition, there is many other configurations like minio.backupXXXX, should it be better if we put them under backup namepace, i.e. backup.XXXX?

### Why is this needed?

_No response_

### Anything else?

_No response_

huanghaoyuanhhy:
Refactors the MinIO-centric configuration into explicit milvus.storage, backup.storage, transfer, backup concurrency, and restore concurrency sections. Replaces crossStorage with transfer.mode (auto/direct/streaming), applies one endpoint-aware transfer policy to backup and restore, and keeps legacy YAML keys and environment variables as deprecated aliases. Updates sample configs and documentation. Testing: go test ./... and git diff --check. Closes #1005.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,11 +48,11 @@ The table below shows which Milvus versions can restore backups created from oth
 Milvus-backup provides command line and API server for usage.
 
 ### Configuration
-In order to use Milvus-Backup, access to Milvus proxy and Minio cluster is required. Configuration settings related to this access can be edited in [backup.yaml](configs/backup.yaml).
+To use Milvus Backup, it must be able to access the Milvus proxy, the storage used by Milvus, and the backup destination. Configure these endpoints in [backup.yaml](configs/backup.yaml).
 
 > [!NOTE]
 >
-> Please ensure that the configuration settings for Minio are accurate. There may be variations in the default value of Minio's configuration depending on how Milvus is deployed, either by docker-compose or k8s.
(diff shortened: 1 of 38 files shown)


huanghaoyuanhhy:
Superseded by the whole-file schema loaders proposed in #1089, which replaced the per-key alias approach prototyped here. Aliases would have allowed mixed-schema files and left every future config change having to understand historical names; the loader now decodes a file as exactly one schema version and translates v1 to v2 in one place.

The work landed as:

- #1095 — the v2 schema and its loader
- #1096 — `config migrate`, which writes an existing v1 file out as v2
- #1102 — the reusable v1-to-v2 translation
- #1104 — version dispatch in the loader, with every consumer moved to v2
- #1105 — the sample configs, and the CI that edits them (open)
- #1106 — more samples, the README and the user guide (open)

The copier rename this PR also carried landed separately as #1091.

Closing in favour of those.

PR #1085, opened 2026-07-17
docs: add agent development rules

huanghaoyuanhhy:
## Summary

Add `AGENTS.md` to record the rules AI coding agents should follow in this repo. Starting with one rule: use testify assertions in tests instead of hand-written `if` checks.

This records an existing convention rather than introducing a new one — 52 of the 53 test files in the repo already use `testify/assert`. Writing it down means agents follow it by default instead of rediscovering it from surrounding code (or not).

`CLAUDE.md` is a symlink to `AGENTS.md`, so Claude Code and the agents that look for `AGENTS.md` (Codex, Cursor) share a single source of truth.

## Changes

- add `AGENTS.md` with the testing rule: `assert` for checks, `require` for preconditions that must hold, PascalCase sub-test names
- add `CLAUDE.md` as a symlink to `AGENTS.md`

/kind improvement

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,48 @@
+# Agent Development Rules
+
+Rules for AI coding agents (Claude Code, Codex, Cursor, ...) working in this repository.
+Human contributors should read [CONTRIBUTING.md](CONTRIBUTING.md) — these rules are additions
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/LGTM

PR #1086, opened 2026-07-17
docs: add etcd-style commit title rule

huanghaoyuanhhy:
## Summary

Record the commit title convention in `AGENTS.md`, following #1085.

Use etcd style — `scope: short description in lowercase` — not Conventional Commits. This matches what the repo already does: of the last 100 non-dependabot commits, 64 use etcd style (`ci:`, `docs:`, `restore:`, `backup:`, `storage:`, ...) against 5 using Conventional Commits (`fix(storage):`, `feat:`). Dependabot's `build(deps):` commits are noted as exempt since they are generated.

The section also records something easy to get wrong: this repo squash-merges, so the **PR title** is what lands in git history, not local commit subjects. #1074 merged 21 commits under a single subject taken from its PR title. An agent that carefully styles its commit subjects but not the PR title has styled the wrong thing.

## Changes

- add a "Commit titles" section to `AGENTS.md` with the etcd-style format, scope list, and examples drawn from real history
- note that Conventional Commits are not used here, and that dependabot commits are exempt
- note that the PR title becomes the squashed commit subject, with ` (#NNNN)` appended

/kind improvement

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,3 +46,27 @@ assert.Len(t, got.GetCollectionBackups(), 1)         // the actual check
 
 Use PascalCase for sub-test names in `t.Run()`, e.g. `t.Run("SecretValueIsMasked", ...)`.
 See `internal/cfg/value_test.go` for a table-driven test in the intended style.
+
+## Commit titles
+
+Use etcd-style subjects — `scope: short description in lowercase`:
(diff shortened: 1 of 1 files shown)


lentitude2tk (approved):
/LGTM

Issue #1089, opened 2026-07-21
[Enhancement]: Introduce a versioned Milvus Backup configuration schema


huanghaoyuanhhy:
## What would you like to be added?

Introduce a versioned, file-level configuration schema so that Milvus client endpoints, storage ownership, authentication, data transfer policy, and task concurrency are explicit without maintaining compatibility as aliases on every individual key.

This proposal expands on #1005. Draft PR #1083 prototypes parts of the new structure, but its per-key alias approach should be replaced by separate whole-file schema loaders.

### Problem

The current configuration mixes unrelated responsibilities and assumes deployment details that are not always true. Maintaining old and new names inside every value resolver would add a second long-term problem:

- Per-key aliases allow mixed-schema files, make precedence difficult to reason about, and require every future config change to understand historical names.
- `minio.storageType` describes the storage used by Milvus, not necessarily MinIO.
- The same `minio` section contains both Milvus storage and backup destination settings.
- Generic storage fields such as `accessKeyID`, `secretAccessKey`, `token`, `gcpCredentialJSON`, and `useIAM` mix several incompatible authentication models. For example, Azure currently treats `accessKeyID` as the storage account name.
- Backup destination fields use a `backup*` prefix instead of a dedicated namespace.
- `minio.crossStorage` is a boolean controlling transfer behavior, but does not describe how data is transferred.
- `backup.parallelism` mixes backup scheduling, restore scheduling, import jobs, and object transfer concurrency.
- `milvus.address` and `milvus.port` are used for both public gRPC and public REST clients, so deployments with separate gRPC and REST endpoints cannot configure them independently.
- `backup.gcPause.address` is actually a Milvus management HTTP endpoint. It is used for both DataCoord GC control and RootCoord EZK export, including encrypted database backup when GC pause is disabled.
- Direct etcd access is required for optional extra metadata backup, but its conditional scope is easy to miss. It is not used by normal backup or restore operations.

This makes the configuration difficult to understand, document, validate, and extend across different Milvus deployments and storage providers.

### Actual Milvus interface usage

| Interface | Current configuration | Client and operations |
| --- | --- | --- |
| Public gRPC | `milvus.address` + `milvus.port` | `NewGrpc`; metadata, collection/database operations, flush, RBAC, legacy import, and other Milvus service RPCs |
[...]

PR #1091, opened 2026-07-21
storage: rename copiers to direct and streaming

huanghaoyuanhhy:
## Summary

`serverCopier` and `remoteCopier` describe the wrong thing. In object storage vocabulary a *server-side copy* is the one the storage service performs itself — which is exactly what `remoteCopier` does. So `serverCopier`, whose "server" means the milvus-backup process, points readers at the other implementation.

This renames both to match what they do, using the same vocabulary as the `transfer.mode` values proposed in #1089 (`direct` / `streaming`), so the types, the log field, and the future config value all read the same.

Pure rename — no behavior change. Split out of #1083 so that PR is left with the configuration structure and endpoint-aware `auto` detection.

## Changes

- `remoteCopier` → `directCopier`: uses the destination storage's native COPY API
- `serverCopier` → `streamingCopier`: streams the object through this process
- `CopyObjectsOpt.CopyByServer` / `CopyPrefixOpt.CopyByServer` → `Streaming`
- log field `copier=remote|server` → `copier=direct|streaming`
- doc comments now say which side moves the bytes

Part of #1089

/kind improvement

### DIFF
--- a/core/backup/coll_dml_task.go
+++ b/core/backup/coll_dml_task.go
@@ -526,11 +526,11 @@ func (dmlt *collDMLTask) backupSegmentData(ctx context.Context, seg *backuppb.Se
 	attrs = append(attrs, insertAttrs...)
 	attrs = append(attrs, deltaAttrs...)
 	opt := storage.CopyObjectsOpt{
-		Src:          dmlt.milvusStorage,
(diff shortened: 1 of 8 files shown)


lentitude2tk (approved):
/LGTM

PR #1093, opened 2026-07-22
storage: abort multipart upload on a detached context

huanghaoyuanhhy:
## Summary

The deferred cleanup in `multiPartCopy` aborted the multipart upload with the
caller's `ctx`. But the most common reason we reach that cleanup is that `ctx` was
already canceled — the user canceled the backup, or one part failed and the
errgroup tore down the shared context. The abort request then fails before it is
even sent, and the already uploaded parts stay in the bucket, taking up space
until a lifecycle rule reaps them.

Run the abort on a context detached from the caller's cancellation, bounded by
its own 30s timeout so the cleanup path cannot hang, and retry it a few times.
`context.WithoutCancel` keeps the ctx values (logging/trace context) and drops
only the cancellation and deadline, which is exactly what a cleanup path wants.

## Changes
- add `MinioClient.abortMultipartUpload`, which aborts on `context.WithoutCancel(ctx)`
  with a 30s timeout and 3 attempts
- call it from the `multiPartCopy` deferred cleanup

/kind improvement

### DIFF
--- a/internal/storage/minio.go
+++ b/internal/storage/minio.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/minio/minio-go/v7"
 	"github.com/minio/minio-go/v7/pkg/credentials"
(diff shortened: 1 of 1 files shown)


lentitude2tk (approved):
/LGTM

PR #1095, opened 2026-07-22
cfg: add the v2 configuration schema and loader

huanghaoyuanhhy:
## Summary

Adds the v2 configuration schema and its loader in a package of their own. The compatibility boundary is the whole file: v2 accepts only v2 keys, environment variables, and `--set` paths, and rejects everything else instead of carrying per-key aliases. A v1 key found in a v2 file reports the v2 key that replaced it.

Nothing dispatches on `configVersion` yet, so v1 remains the only schema the application loads. The complete v2 sample lives in the package testdata rather than `configs/`, so no user-facing file claims v2 is loadable.

Part of #1089

## Changes

- Move the parameter machinery both versions resolve values with into `internal/cfg/param` (typed values, source precedence, secret masking, config table). v1 keeps its behaviour and public shape through type aliases, so no consumer changed.
- Add `param.List` for list-valued parameters: a YAML list in a config file, comma separated as an environment variable or override.
- Add `internal/cfg/v2` with the v2 schema:
  - `milvus.grpc`, `milvus.rest`, and `milvus.management` are modelled independently, so split-endpoint deployments are configurable.
  - `milvus.storage` and `backup.storage` share one provider-discriminated shape with a typed `auth` block (`static`, `sharedKey`, `serviceAccount`, `iam`, `default`). Azure's `accountName` belongs to the storage identity, not the credential.
  - `transfer` replaces `minio.crossStorage` and carries the object transfer concurrency that `backup.parallelism.copydata` used to mix with task scheduling.
  - `backup.concurrency` and `restore.concurrency` split the remaining `backup.parallelism` knobs by the operation they belong to.
  - `milvus.etcd.endpoints` becomes a YAML list.
  - `milvus.grpc.tlsMode` becomes `disabled` / `server` / `mutual`. The v1 numbers were copied from the Milvus *server side* parameter `common.security.tlsMode`, while milvus-backup is a client; the official Go SDK does not use them either. The field keeps its place, and the migration hint names the mode each number meant.
  - `http` becomes `server` (the unused `http.enabled` is dropped), `cloud` becomes `zillizCloud`, `log.file.filename`/`maxSize` become `path`/`maxSizeMiB`, and `milvus.rpcChannelName` moves under `milvus.replicate`.
- Reject unknown config file keys and `--set` paths before resolving, reporting all of them at once. Known v1 names are reported with the v2 name that replaced them.
- Validate the resolved configuration, joining every failure into one error: provider and authentication type compatibility, credentials that do not apply to the selected authentication type, ports, endpoint URLs, and positive concurrency values.

## Notes

Two intentional differences from v1, both at the version boundary:

- `backup.storage` inherits from `milvus.storage`, except `rootPath`, which keeps its own `backup` default. In v1 an unset `backupRootPath` inherited `minio.rootPath`, putting backup data under the Milvus root path.
- `tlsMode: mutual` without an mTLS key pair is now an error. v1 warns and silently downgrades to server TLS, which its own comment marked for removal.

v1 environment variables are not consulted by the v2 loader, matching the "they are not cross-version aliases" rule. Whether a v1 variable set alongside a v2 file should warn is left to the change that adds version dispatch (question 4 in #1089).

/kind improvement


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

PR #1096, opened 2026-07-23
cfg: add config migrate command to translate v1 config to v2

huanghaoyuanhhy:
## Summary

Adds `milvus-backup config migrate`, which reads a v1 configuration file and writes a complete v2 file (the migration step of #1089). Migration renames keys and environment variables rather than relocating values: a secret an operator keeps in a v1 environment variable is left out of the file and reported with the v2 variable to set instead, so nothing secret is written to disk.

The v2 loader added in #1095 already rejects v1 keys with the v2 name that replaced them; this command performs that rewrite automatically, including the value transforms a flat key rename cannot express.

```
$ milvus-backup config migrate --config backup.yaml -o backup-v2.yaml
# the v2 file goes to -o (or stdout); the migration report goes to stderr
```

## Changes

- `cmd/config`: new `config` command group with `migrate`. Reads the v1 file via `--config`, writes v2 to stdout or `--output`; `--force` guards overwrite, `--strict` fails when the result is not valid v2. The report (warnings + env renames) is written to stderr so stdout stays a clean document.
- `internal/cfg/migrate`: maps a resolved v1 config to a v2 config plus a report. Handles the value transforms — `tlsMode` `0/1/2` → `disabled/server/mutual` (mutual without an mTLS pair downgrades to server, as v1 did silently), provider aliases (`ali`/`tc` → `aliyun`/`tencent`), the provider-aware auth reshape (Azure account name/key, GCP service account, IAM), `crossStorage` → `transfer.mode`, and the `http.enabled` drop.
- Secrets in v1 environment variables are deferred, not written: the field keeps a placeholder with an inline comment naming the v2 variable, and the report lists every v1 variable that needs renaming. An unset backup secret that only inherited an env-supplied primary secret is deferred too, so it does not leak into the file.
- `internal/cfg/v2/render.go`: serializes a resolved v2 config back to an ordered YAML file, omitting credential fields that do not apply to the selected provider or auth type so the output loads.
- `internal/cfg/param`: adds `YAMLValue` and `UseDefault`/`Defaults`, used by the renderer and the migration baseline; exports `v2.Migration` for the env rename report.
- Routes the startup version banner and the `config:` line to stderr, so a data-producing command such as `config migrate` keeps stdout a clean document.

## Notes

- Nothing dispatches on `configVersion` yet, so this does not change how the application loads configuration; it only produces v2 files. Version dispatch, `check config`, and the shipped v2 sample are follow-ups.
- `minio.crossStorage: false` maps to `transfer.mode: auto`, which is not behaviour-identical to v1 when the milvus and backup storage differ; the migration warns in that case (per #1089).

Part of #1089

/kind improvement

### DIFF
--- a/cmd/cmd.go
+++ b/cmd/cmd.go
@@ -5,6 +5,7 @@ import (
 	"os"
 
 	"github.com/zilliztech/milvus-backup/cmd/check"
+	"github.com/zilliztech/milvus-backup/cmd/config"
 	"github.com/zilliztech/milvus-backup/cmd/create"
 	"github.com/zilliztech/milvus-backup/cmd/del"
 	"github.com/zilliztech/milvus-backup/cmd/get"
@@ -22,6 +23,7 @@ func Execute() {
 	cmd := root.NewCmd(&opt)
 
 	cmd.AddCommand(check.NewCmd(&opt))
+	cmd.AddCommand(config.NewCmd(&opt))
 	cmd.AddCommand(create.NewCmd(&opt))
 	cmd.AddCommand(del.NewCmd(&opt))
 	cmd.AddCommand(get.NewCmd(&opt))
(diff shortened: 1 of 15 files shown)


lentitude2tk (approved):
/LGTM

PR #1097, opened 2026-07-23
cfg: warn on unknown v2 config keys instead of failing

huanghaoyuanhhy:
## Summary

The v2 config loader rejected the whole file when it carried an unknown key —
a misspelling, a leftover v1 name, or an extra field. That makes an otherwise
valid config unloadable over a single stray line, which is heavier than the
mistake warrants: unknown keys are never applied anyway, since resolving only
reads declared keys.

This relaxes the unknown-key check from a hard error to a warning. Each unknown
config-file key and `--set` path is logged via `log.Warn` and then ignored, and
loading proceeds. The warning still names the key and, for v1 leftovers, the v2
key to use instead, so the diagnostic is unchanged — only its severity is.

Version detection (`configVersion`) and value validation (port ranges, provider
enums, required credentials, ...) stay hard errors: the former drives v1/v2
dispatch, and the latter would otherwise let an invalid value flow through to a
confusing runtime failure.

## Changes

- `LoadFrom`: log unknown-key warnings and continue instead of aborting the load
- `checkKeys` now returns `[]string` warnings; `unknownKeyErr` becomes
  `unknownKeyWarning` and its messages note the key is being ignored
- Tests: assert the warnings on `checkKeys`'s return value, and add a case that
  an unknown key is ignored so the affected field keeps its default

/kind improvement

### DIFF
--- a/internal/cfg/v2/load.go
+++ b/internal/cfg/v2/load.go
@@ -1,17 +1,18 @@
 package v2
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 
 	"github.com/zilliztech/milvus-backup/internal/cfg/param"
+	"github.com/zilliztech/milvus-backup/internal/log"
 )
 
 // Load loads a v2 configuration from yaml + overrides + env.
 //
 // precedence: overrides (--set) > env > config file > default, among the names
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/LGTM

PR #1098, opened 2026-07-24
config: add config show and deprecate check config

huanghaoyuanhhy:
## Summary

Add a `config show` command that prints the resolved configuration table, and turn the old `check config` into a deprecated alias.

The configuration dump only printed the resolved values — it never connected to anything, so it did not belong under `check` (whose job is verifying connectivity). The `config` group's own description is "inspect and migrate", but until now it only had `migrate`. This wires up the "inspect" half as `config show`.

`check config` is kept as a deprecated stub: running it prints a deprecation notice pointing at `config show` and nothing else, and it is hidden from the `check` help listing. The stub is slated for removal in a release after 0.6.0. The `check` command still prints its effective configuration inline before connecting, so nothing is lost there.

## Changes

- Add `config show`, printing all configuration parameters with their values and sources
- Deprecate `check config`: it now only emits a notice redirecting to `config show` (via cobra's `Deprecated`, which also hides it from help), marked for removal in a release after 0.6.0
- Update `docs/FAQ.md` to reference `config show`

/kind improvement

### DIFF
--- a/cmd/check/check.go
+++ b/cmd/check/check.go
@@ -64,7 +64,7 @@ func NewCmd(opt *root.Options) *cobra.Command {
 		},
 	}
 
-	cmd.AddCommand(newConfigCmd(opt))
+	cmd.AddCommand(newConfigCmd())
 
 	return cmd
 }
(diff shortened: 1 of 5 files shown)


lentitude2tk (approved):
/LGTM

PR #1101, opened 2026-07-24
test: use testify assertions in l0compact tests

huanghaoyuanhhy:
## Summary

The l0compact tests were written with hand-rolled `if` checks followed by `t.Fatalf` /
`t.Errorf`, which the repo's agent rules explicitly call out as the wrong style. This
converts them to testify assertions so a failure prints an expected-vs-actual diff instead
of an ad-hoc message.

Test-only change, no production code touched.

## Changes

- Convert every hand-written check in `internal/l0compact/` (deltalog, parquet, insertpk, envelope, pk, version) and `cmd/l0compact/` to `assert` / `require`
- Use `require` for preconditions (setup writers, values about to be dereferenced) and `assert` for the actual checks, so one run reports every mismatch
- Replace compound conditions like `len(out) != 1 || out[0].PK.Int != 42 || out[0].Ts != 100` with a single `assert.Equal` on the whole value
- Stop discarding errors in `deltalog_test.go` — `WriteParquetPKTs` and `BuildV1Envelope` results are now checked with `require.NoError`

`core/l0compact/` already used testify and is unchanged.

/kind improvement

### DIFF
--- a/cmd/l0compact/l0compact_test.go
+++ b/cmd/l0compact/l0compact_test.go
@@ -1,29 +1,23 @@
 package l0compact
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
 
 func TestValidateRejectsBadNames(t *testing.T) {
 	// --output that would escape or nest outside a normal backup dir.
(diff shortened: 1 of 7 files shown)


lentitude2tk (approved):
/LGTM

PR #1102, opened 2026-07-24
cfg: extract v1 to v2 translation for the config loader

huanghaoyuanhhy:
## Summary

Prepares the config loader for dispatching on the schema version: a v1 file will be decoded with the v1 schema and then translated into v2, so downstream code only ever sees a v2 config. That needs the v1 to v2 mapping `Migrate` already implements, but `Migrate` cannot be reused as is — it withholds a secret that reached v1 through an environment variable, so the file it emits does not bake the secret in, while a running process needs that value to connect.

## Changes

- Extract the v1 to v2 field mapping out of `Migrate` into a shared `translate`.
- Add `Translate`, the loader-facing entry point: it carries every value into the result, including one that came from an environment variable, and returns the v2 validation error instead of collecting it for a human to read.
- Make `*Report` nil-safe. A nil report selects the loader path, so the report-only branches turn into no-ops rather than growing a second set of call sites.
- Add tests for the env-secret difference, the value decisions `Migrate` would warn about (mutual TLS downgrade, inherited backup root path, crossStorage), the validation error path, and that both entry points agree when there is no secret to withhold.

`Migrate` and the `config migrate` command keep their behavior.

Part of #1089

### DIFF
--- a/internal/cfg/migrate/migrate.go
+++ b/internal/cfg/migrate/migrate.go
@@ -1,13 +1,18 @@
 // Package migrate translates a resolved v1 configuration into a v2
-// configuration plus a report of everything that needs a human's attention.
+// configuration. Two entry points share one field mapping: Migrate, which also
+// returns a report of everything that needs a human's attention and backs the
+// `config migrate` command, and Translate, which the config loader uses to run
+// an existing v1 file through the v2 schema.
 //
 // Migration renames configuration keys and environment variables; it does not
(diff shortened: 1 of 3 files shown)


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

lentitude2tk (approved):
/LGTM

PR #1104, opened 2026-07-25
cfg: load either schema version and run the program on v2

huanghaoyuanhhy:
## Summary

The v2 schema landed alongside v1 but nothing ran on it: the loader resolved v1 names and every consumer read the v1 config. This points the program at v2 and keeps existing v1 files working by translating them, so the switch is atomic — there is no half-migrated state where some components read one schema and some read the other.

## Changes

- Add `internal/cfg/loader`, which dispatches on the `configVersion` discriminator: a v2 file loads directly, and a file without the key predates it, so it is resolved with the v1 schema — including the v1 environment variables — and translated with `migrate.Translate`. The file is read once and decoded as exactly one schema version. With no config file there is nothing to be backward compatible with, so overrides and env name v2 parameters.
- Take v2 types in the adapters. The storage factory switches on `auth.type` instead of inferring credentials from the provider, a `useIAM` flag and whether a GCP credential file happened to be set; the provider switch drops the v1 spelling aliases (`ali`, `alibaba`, `alicloud`, `tc`), which are folded into the canonical name while the config is loaded. `NewGrpc` reads the named TLS modes, and `NewRestful` prefers a configured REST endpoint over deriving one from the gRPC connection.
- Stop importing a config schema in `internal/log`: the caller maps its own configuration onto `log.Config`, which is what lets a schema package log.
- Replace `minio.crossStorage` with `storage.UseStreaming` over `transfer.mode`. `auto` keeps what v1 did for differing providers, and additionally streams between two deployments of the same provider — which v1 attempted to storage-side copy and could not. The local `crossStorage` names follow the config.
- Convert every consumer under `cmd/` and `core/` to the v2 field paths, and the affected tests along with them.

Two silent fallbacks are gone. Mutual TLS without a readable client key pair is now an error rather than a quiet downgrade to server TLS — validation rejects a v2 config that asks for it and translation settles the v1 downgrade beforehand, so nothing reaches the client. `gcpnative` says it needs a service account credentials file instead of failing on an empty path.

`configs/backup.yaml` deliberately stays v1: CI injects v1 keys into it with `yq`, so leaving it alone keeps the E2E matrix exercising the translation path. Converting it — and the `yq` keys with it — belongs in its own PR.

Part of #1089

### DIFF
--- a/cmd/check/check.go
+++ b/cmd/check/check.go
@@ -9,13 +9,13 @@ import (
 
 	"github.com/zilliztech/milvus-backup/cmd/root"
 	"github.com/zilliztech/milvus-backup/core/check"
-	"github.com/zilliztech/milvus-backup/internal/cfg"
+	v2 "github.com/zilliztech/milvus-backup/internal/cfg/v2"
 	"github.com/zilliztech/milvus-backup/internal/client/milvus"
 	"github.com/zilliztech/milvus-backup/internal/storage"
 )
 
 // writeConfig prints a labeled table of the effective configuration to w.
-func writeConfig(w io.Writer, c *cfg.Config) error {
+func writeConfig(w io.Writer, c *v2.Config) error {
 	if _, err := io.WriteString(w, "Configuration:\n"); err != nil {
 		return fmt.Errorf("check: write config header: %w", err)
 	}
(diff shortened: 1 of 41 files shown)


lentitude2tk (approved):
/LGTM

lentitude2tk (approved):
LGTM

PR #1105, opened 2026-07-25
cfg: ship the sample configs as v2

huanghaoyuanhhy:
## Summary

The loader reads either schema version, but the files shipped with the repo were still v1, so every run started by translating them and warning about it. This writes them as v2.

CI edits `configs/backup.yaml` with `yq` and injects credentials through the environment, both of which name keys. A v2 file with v1 keys written into it loads with those keys **ignored** — a test running against the wrong bucket rather than a failure — so the workflows have to move in the same commit.

## Changes

- Rewrite `configs/backup.yaml` and `configs/backup-azure.yaml` as v2, keeping the explanatory comments and adding the ones the new shape needs (auth types, storage inheritance, transfer modes).
- Comment out every credential a given auth type does not use. v2 rejects a credential its type cannot use rather than ignoring it, so listing them all as empty strings does not load — the sample now shows which field belongs to which `auth.type`.
- Update the `yq` paths in `main.yaml`, `nightly.yaml`, `perf.yaml` and `azure.yaml` to their v2 spellings, e.g. `.minio.bucketName` → `.milvus.storage.bucketName` and `.minio.backupRootPath` → `.backup.storage.rootPath`.
- Map `.minio.crossStorage = true` to `.transfer.mode = "streaming"` in the cross-storage job, which backs up between two MinIO deployments on different ports.
- Pass the azure credentials as `MILVUS_STORAGE_ACCOUNT_NAME` / `MILVUS_STORAGE_AUTH_ACCOUNT_KEY` (and the backup-side pair) instead of overloading `MINIO_ACCESS_KEY` / `MINIO_SECRET_KEY`, which is how v1 spelled an Azure account name and key.

## Verification

Both sample files were loaded with `config show` to confirm they resolve. The `main.yaml` upstream-test block was replayed against the new `configs/backup.yaml` — every injected value lands on the intended v2 parameter with `SOURCE=config`, which is the failure mode this change is guarding against. The azure file was loaded with the four new environment variables set, confirming the credentials resolve with `SOURCE=env`.

Docs still describe v1 keys (`docs/user_guide/env_variables.md` is a full v1 variable table, and `cross_storage.md` is written around `crossStorage`); that is its own change.

Part of #1089

### DIFF
--- a/.github/workflows/azure.yaml
+++ b/.github/workflows/azure.yaml
@@ -83,8 +83,8 @@ jobs:
         run: |
           # Update backup config with unique paths
           cp backup-azure.yaml backup.yaml
-          yq -i ".minio.rootPath = \"${{ steps.paths.outputs.milvus_root }}\"" backup.yaml
-          yq -i ".minio.backupRootPath = \"${{ steps.paths.outputs.backup_root }}\"" backup.yaml
+          yq -i ".milvus.storage.rootPath = \"${{ steps.paths.outputs.milvus_root }}\"" backup.yaml
+          yq -i ".backup.storage.rootPath = \"${{ steps.paths.outputs.backup_root }}\"" backup.yaml
 
       - name: Milvus deploy
         timeout-minutes: 15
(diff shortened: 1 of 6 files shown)


lentitude2tk (approved):
/LGTM

PR #1106, opened 2026-07-25
docs: describe the v2 configuration

huanghaoyuanhhy:
> Stacked on #1105 — review the top commit. The diff shown here includes #1105 until that merges.

## Summary

The samples and the docs were still written for the v1 schema: README named a `minio` section that no longer exists, `env_variables.md` listed the v1 variables, and `cross_storage.md` was built around `minio.crossStorage`. This rewrites them for v2 and adds samples for the deployments people actually ask about.

## Changes

- Add `configs/backup-s3.yaml`, `backup-gcp.yaml`, `backup-local.yaml` and `backup-iam.yaml`, covering backups on S3, on Google Cloud Storage, on a local disk, and an S3 setup that authenticates with an instance role and keeps no secret in the file. Each names only what differs from `milvus.storage`, so the samples demonstrate the inheritance instead of repeating a backend twice.
- Trim `backup-azure.yaml` to the same shape. It was carrying a `replicate` channel it never uses — that section configures the CDC message channel, which has nothing to do with Azure storage — along with a page of restated defaults inherited from its v1 original. The resolved configuration is unchanged; every removed key already held its default.
- Add `make check-configs`, which builds the binary and loads every file in `configs/` through it, and run it in CI as its own job. A sample is what people copy, and a broken one is otherwise only found by the person copying it. It is also why the credentials in the samples are placeholders rather than empty strings — a required field left empty does not load, which was true of `backup-azure.yaml`.
- Rewrite the README configuration section around the v2 layout, add a table of the example files, and document `config migrate` for anyone carrying an older file forward.
- Regenerate `env_variables.md` from the schema, so the table matches what the code declares, and describe which credentials each `auth.type` accepts.
- Rename `cross_storage.md` to `transfer.md`, after the `transfer` section it now describes. `crossStorage` was a v1 flag, and `auto` covers the case it had to be set by hand for — the FAQ says so for anyone arriving from the old name.
- Fix the v1 key names in `FAQ.md`.

## Verification

`make check-configs` passes over all six samples, and fails as intended when a sample is broken. Every YAML block in the docs was extracted and loaded, and every relative link in the touched markdown was resolved against the tree. The doc snippets carry `configVersion`, because a snippet without it is read as v1 — the v2 keys in it would be ignored rather than rejected, which is the same silent-misconfiguration failure #1105 guards against.

Part of #1089. The remaining work under that issue is retiring the v1 schema, which is its own change.


lentitude2tk (approved):
/LGTM

PR #1108, opened 2026-07-25
storage: log the resolved transfer mode when a task starts

### LINKED ISSUE #993: [Feature]: Replace crossStorage with smarter auto-detection and clearer copyMode config
Part of #637 (Overly Complex Configuration File)

## Is your feature request related to a problem? Please describe.

The current `minio.crossStorage` configuration is confusing and hard to understand for users:

1. **The name describes the scenario, not the behavior.** `crossStorage` sounds like "are you doing cross-storage backup?" but it actually controls **how data is copied** — via the storage system's native COPY API (`false`) or proxied through the backup server process (`true`). Users have to read a dedicated FAQ just to understand what it means (`docs/user_guide/cross_storage.md`).

2. **Manual configuration is needed for same-type-different-instance scenarios.** Auto-detection (`core/backup/task.go:111`) only compares `Provider` — if both are `minio` but at different addresses (Minio A → Minio B), the user must manually set `crossStorage: true`. This is the most confusing case: "both are Minio, why do I need crossStorage?"

3. **Wrong config section.** `crossStorage` is under the `minio` section, but it's not a MinIO setting — it's a backup behavior setting.

4. **The FAQ in the docs is itself evidence of a naming failure.** If a config option needs a dedicated FAQ entry explaining "what does this mean", the name isn't working.

## Describe the solution you'd like.

This is a **breaking change** proposal:

### 1. Smarter auto-detection (P0) — eliminate manual configuration

Extend auto-detection to compare `address` and `port` in addition to `provider`:

```go
// current: only compares provider
if backupProvider != milvusProvider { copyByServer = true }

// proposed: also compare address/port
if backupProvider != milvusProvider ||
   backupAddress != milvusAddress ||
   backupPort != milvusPort {
    copyByServer = true
}
```

This covers the Minio A → Minio B scenario automatically. Users almost never need to set this manually.

### 2. Rename the config (P1) — name should describe behavior, not scenario

Replace `minio.crossStorage` with an enum-based `copyMode`:

```yaml
backup:
  copyMode: "auto"     # auto-detect (default, recommended)
  # copyMode: "direct" # force storage COPY API
  # copyMode: "proxy"  # force server-side proxy transfer
```

- `auto`: current implicit behavior, picks the right mode automatically
- `direct`: uses storage's native COPY API (fast, requires both ends accessible to each other)
- `proxy`: data flows through backup server (slower, works across any storage combination)

### 3. Move config from `minio` to `backup` section (P2)

`crossStorage` is not a storage configuration — it's a backup operation behavior. It belongs under the `backup` section.

### 4. Log the actual copy mode at task start (P3)

```
INFO  copy mode: proxy (auto-detected: backup and milvus storage have different addresses)
INFO  copy mode: direct (same storage, using native COPY API)
```

## Describe an alternate solution.

A minimal non-breaking approach: keep `crossStorage` as-is but improve auto-detection (P0) so users rarely encounter it. Add deprecation warnings pointing to the new config name.

### Code references
- Config definition: `internal/cfg/cfg.go:234,277`
- Auto-detection: `core/backup/task.go:110-113`, `core/check/task.go:139-150`
- Copier selection: `internal/storage/copier.go:126-137`
- Restore skip logic: `core/restore/coll_task.go:336-343`
- Backup copy: `core/backup/coll_dml_task.go:474-483`
- User docs: `docs/user_guide/cross_storage.md`

### Migration path
- Support both old (`minio.crossStorage`) and new config for one release cycle with a deprecation warning
- Remove the old config in the following release

huanghaoyuanhhy:
## Summary

Backup and restore resolve `transfer.mode` into a decision — stream the objects through this process, or ask the storage backend to copy them — and then never report it. With the mode left at `auto` that decision is not one the operator made, so a backup running at network speed instead of backend speed leaves nothing in the log to explain itself. This is the last open item of #993.

## Changes

- Log the resolved transfer mode in `backup.NewTask` and `restore.NewTask`, next to where the decision is made. The configured mode is logged alongside the resolved boolean, so an `auto` that resolved to streaming reads differently from a `streaming` that was asked for — with only the boolean, the interesting case and the ordinary one look identical.
- Name the mode in the line `check` already emits, rather than adding a second one beside it. All three paths now carry the same `transfer_mode` field, so one grep covers them.
- Hoist `streaming` out of the `restore.Task` literal into a local, since it now has a second reader. The value and the comment explaining that it is decided once are unchanged.

## Notes

Two places where this differs from the letter of #993, both deliberate:

- The issue asks for the mode "at task start". `check` decides at the copy step rather than at task start, because that is where the two backends are known, so its line stays there — moving it would report a decision the task has not made yet.
- The issue's other three items landed in #1104 under different names than it proposed: `copyMode: auto/direct/proxy` under `backup` became `transfer.mode: auto/direct/streaming` at the top level, because `check` and `migrate` share the policy and it is not backup-specific. The behavior asked for is the behavior shipped.

## Verification

`go test --race ./...` and `golangci-lint run ./...` pass. The line was confirmed to actually emit by constructing a restore task against two mock backends and reading the output — worth checking, because the default logger in this package discards everything until `InitLogger` installs a sink, so a compiling log call proves nothing on its own.

No test asserts on the log. `TestUseStreaming` already covers which mode gets chosen, and the repository has no `zaptest/observer` precedent to assert on log output.

Fixes #993

### DIFF
--- a/core/backup/task.go
+++ b/core/backup/task.go
@@ -108,6 +108,9 @@ func NewTask(args TaskArgs) (*Task, error) {
 
 	streaming := storage.UseStreaming(args.Params.Transfer.Mode.Val,
 		args.MilvusStorage.Config(), args.BackupStorage.Config())
+	logger.Info("resolved transfer mode",
+		zap.String("transfer_mode", args.Params.Transfer.Mode.Val),
+		zap.Bool("streaming", streaming))
 
 	mb := newMetaBuilder(args.TaskID, args.Option.BackupName)
 	err := args.TaskMgr.AddBackupTask(args.TaskID, args.Option.BackupName)
--- a/core/check/task.go
+++ b/core/check/task.go
@@ -130,7 +130,10 @@ func (t *Task) checkWriteAndCopy(ctx context.Context) error {
 	t.logger.Info("copy from milvus storage to backup storage")
 	streaming := storage.UseStreaming(t.params.Transfer.Mode.Val,
 		t.milvusStorage.Config(), t.backupStorage.Config())
-	t.logger.Info("try to copy", zap.Bool("streaming", streaming), zap.String("dest_key", destKey))
+	t.logger.Info("try to copy",
+		zap.String("transfer_mode", t.params.Transfer.Mode.Val),
+		zap.Bool("streaming", streaming),
+		zap.String("dest_key", destKey))
 	opt := storage.CopyPrefixOpt{
 		Src:        t.milvusStorage,
 		Dest:       t.backupStorage,
--- a/core/restore/task.go
+++ b/core/restore/task.go
@@ -168,11 +168,16 @@ func NewTask(args TaskArgs) (*Task, error) {
 
 	args.TaskMgr.AddRestoreTask(args.TaskID)
 
+	streaming := storage.UseStreaming(args.Params.Transfer.Mode.Val,
+		args.BackupStorage.Config(), args.MilvusStorage.Config())
+	logger.Info("resolved transfer mode",
+		zap.String("transfer_mode", args.Params.Transfer.Mode.Val),
+		zap.Bool("streaming", streaming))
+
 	return &Task{
 		args: args,
 
-		streaming: storage.UseStreaming(args.Params.Transfer.Mode.Val,
-			args.BackupStorage.Config(), args.MilvusStorage.Config()),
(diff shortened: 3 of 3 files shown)


lentitude2tk (approved):
/LGTM

PR #1109, opened 2026-07-25
cmd: reject the flags deprecated in 0.5

huanghaoyuanhhy:
## Summary

0.5 accepted these flags with a deprecation warning; 0.6 stops honoring them. Each one stays
registered but hidden from help, so an outdated command line is answered by the flag that has to
change rather than by cobra's bare `unknown flag`:

```console
$ milvus-backup create --colls c1
Error: --colls was removed in 0.6, use --filter instead
```

The registration is not what makes this safe — cobra already fails on an unknown flag and exits 1,
so the stubs only buy the message. Deleting them in a later release degrades that message without
letting anything through, which is why the removal is left as a TODO rather than done here.

One replacement is not a drop-in swap. On `restore`, `--filter` matches the collection name **after**
`--suffix` or `--rename` is applied, while the removed `--collections` matched the name in the backup.
So `restore -c hello_milvus -s _recover` becomes `restore --filter hello_milvus_recover -s _recover`;
filtering on the pre-rename name matches nothing and restores nothing. This is now stated in
`restore --help` and the README, and the CI job that relied on `-c` is rewritten accordingly.

## Changes

- Add `cmd/flags`, where one table per command drives registration, hiding and the error message, so
  the advice in help and the advice in the error cannot drift apart.
- Reject the removed flags in `create` (`--colls`, `--databases`, `--database_collections`, `--force`,
  `--meta_only`), `restore` (`--collections`, `--databases`, `--database_collections`,
  `--restore_index`), `get` (`--detail`) and `list` (`--collection`).
- Drop the code that existed only to serve them: six `*ToFilter` / `*ToBackupFilter` helpers, the
  `force` and `meta_only` branches of `toStrategy`, and six mutual-exclusion checks in `restore`
  validation. `Plan.BackupFilter` stays, since the HTTP API still sets it.
- Document the post-rename `--filter` semantics in `restore --help` and a new README section, and fix
  the two callers that still passed removed flags (`.github/workflows/main.yaml`,
  `docs/user_guide/e2e_demo_cli.md`).

Part of #695

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -375,8 +375,10 @@ jobs:
         timeout-minutes: 5
         shell: bash
         run: |
-          ./milvus-backup restore -n my_backup -c hello_milvus -s _recover
-          ./milvus-backup restore -n my_backup -c hello_milvus2 -s _recover
+          # --filter matches the collection name after -s is applied, so it
+          # names hello_milvus_recover rather than hello_milvus.
+          ./milvus-backup restore -n my_backup --filter hello_milvus_recover -s _recover
+          ./milvus-backup restore -n my_backup --filter hello_milvus2_recover -s _recover
       - name: Verify data
         timeout-minutes: 5
         shell: bash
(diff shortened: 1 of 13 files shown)


lentitude2tk (approved):
/LGTM

PR #1111, opened 2026-07-25
log: write the bootstrap logger to stderr

### LINKED ISSUE #1110: [Bug]: configuration warnings are discarded before the logger is initialized
### Milvus-Backup Version

`main` at c3ca987 (after v0.5.16, unreleased)

### Can you reproduce this issue on the latest version?

No, it only happens on my current version

### Current Behavior

Warnings raised while the configuration file is being read never reach the user. They are produced, and then discarded.

Two of them exist today, and both are silent.

**A misspelled key is detected and the warning is dropped.** With this file:

```yaml
configVersion: v2
milvus:
  storage:
    provider: minio
    bucketNaem: typo-here
    bukcetName: another-typo
```

```console
$ milvus-backup config show --config typo.yaml
...
Milvus.Storage.BucketName            a-bucket                      default
$ echo $?
0
```

Nothing is printed on stdout or stderr about either misspelling, the run succeeds, and the bucket silently falls back to its default. `checkKeys` does detect both keys and does build a warning for each — the warnings simply go nowhere.

**The v1 deprecation warning is dropped.** Loading a file written in the v1 schema translates it to v2 and is supposed to say so once per run. With a v1 file:

```console
$ milvus-backup config show --config v1.yaml 2>&1 | grep -c "translated it to v2"
0
```

The translation does happen — a `minio.bucketName` set in the v1 file arrives as `milvus.storage.bucketName` — but nothing tells the operator that their file is on the old schema.

### Expected Behavior

Both warnings reach the user. A key that was ignored because it is misspelled is the case where a silent success is most expensive: the backup runs, against a default the operator did not choose.

### Steps To Reproduce

```markdown
1. Write the v2 file above, with a misspelled key under `milvus.storage`.
2. Run `milvus-backup config show --config typo.yaml`, capturing both stdout and stderr.
3. Observe that no warning names the misspelled key, that the exit code is 0, and that the resolved value is the default.
```

### Environment

```markdown
darwin/arm64, go1.26.4, built from source at c3ca987
```

### Anything else?

The cause is an ordering one, and it is not speculation — reverting it makes both warnings appear.

`cmd/root/root.go` loads the configuration before it can initialize logging from that configuration:

```go
params, err := loader.Load(o.Config, overrides)   // both warnings are raised in here
...
log.InitLogger(logConfig(&params.Log))            // the first sink is installed here
```

Until `InitLogger` runs, the global logger is the one from `newStdLogger`, which builds its configuration as `&Config{Level: "info", File: FileLogConfig{}}` — no file name, and `Console` left false. `initLogger` only appends a sink for a file name or for console, so that logger has no sinks at all and discards everything written to it.

This is reachable only on `main`: the released loader raises no warnings while reading, so there was nothing to lose before the v2 schema landed.

Related: #1089, and #1107 quotes the deprecation warning as something operators will see.

huanghaoyuanhhy:
## Summary

The configuration file is read before the log settings inside it can be applied, so anything the loader has to say goes through the logger `newStdLogger` builds. That logger was given a `Config` with no file name and `Console` left false, and `initLogger` only appends a sink for one of those — so it had no sinks and discarded everything written to it.

Two warnings went that way, and both matter more than most: the one naming a config key that was ignored because it is misspelled, and the one saying a v1 file was translated to v2. The first is the expensive one. `checkKeys` detects the misspelling and builds the warning; the backup then runs against a default the operator did not choose, exit code 0, nothing printed.

## Changes

- Build the bootstrap logger on stderr instead of on sinks derived from a configuration that has not been read yet. stderr rather than stdout because a command's own output goes to stdout — `config migrate` writes the generated v2 file there, and a warning landing in it would corrupt the file being generated.
- Drop the `zap.AddCallerSkip(1)` that logger was built with. The package-level `Warn` and friends already add one for their own frame, and the logger `InitLogger` builds adds none, so the extra skip attributed every warning to the caller of the function that logged it — the unknown-key warning was reported at `loader/load.go:53` rather than `v2/load.go:37`. Invisible until now, since nothing this logger wrote was reaching anyone.
- Add `internal/log/log_test.go`, which had no tests. Two cases: the logger writes to stderr, and it reports the line that logged.

## Verification

Reverting either half of the fix fails the corresponding test — checked, because the first version of the call-site test passed either way. It asserted only the file name, and an extra caller skip lands on the test helper, which is in that same file; it now asserts the exact line.

End to end, on a v2 file with two misspelled keys, and on a v1 file:

```console
$ milvus-backup config show --config typo.yaml 2>&1 >/dev/null
[WARN] [v2/load.go:37] ["cfg: unknown v2 config file key \"milvus.storage.bucketnaem\", ignoring it"]
[WARN] [v2/load.go:37] ["cfg: unknown v2 config file key \"milvus.storage.bukcetname\", ignoring it"]

$ milvus-backup config show --config v1.yaml 2>&1 >/dev/null
[WARN] [loader/load.go:77] ["cfg: loaded a v1 configuration and translated it to v2; run `milvus-backup config migrate` to convert the file itself"] [path=v1.yaml]
```

Both now report the line that actually logged them. stdout is unchanged in both runs, and a clean v2 config produces no new output. `go test --race ./...` and `golangci-lint run ./...` pass.

## Notes

Warnings raised before `InitLogger` go to stderr and not to a configured log file, since the file is named by the configuration still being read. For a deployment logging to a file, these two lines land in stderr — container logs — rather than that file. Buffering them for replay after initialization would fix that, and is worth doing if it turns out to matter; it is more machinery than a startup warning seemed to justify here.

Fixes #1110

### DIFF
--- a/internal/log/log.go
+++ b/internal/log/log.go
@@ -125,9 +125,21 @@ func initFileLog(cfg *FileLogConfig) (*lumberjack.Logger, error) {
 	}, nil
 }
 
+// newStdLogger builds the logger the package starts with, the one in use until
+// InitLogger replaces it with the configured one.
+//
+// It writes to stderr rather than going through initLogger, which builds its
+// sinks from the file and console settings and so would have none to build from
+// here. Anything logged while the configuration is still being read — an
+// unknown key, a deprecated schema version — would otherwise be discarded,
+// which is the opposite of what a warning raised at that point is for. stderr
+// keeps it clear of the stdout a command writes its own output to.
+//
+// No caller skip: the package-level Warn and friends already add one for their
+// own frame, and the logger InitLogger builds adds none.
 func newStdLogger() (*zap.Logger, *ZapProperties) {
 	conf := &Config{Level: "info", File: FileLogConfig{}}
-	lg, r, _ := initLogger(conf, zap.AddCallerSkip(1))
+	lg, r, _ := InitLoggerWithWriteSyncer(conf, zapcore.AddSync(os.Stderr))
 	return lg, r
 }
 
--- a/internal/log/log_test.go
+++ b/internal/log/log_test.go
@@ -1,0 +1,65 @@
+package log
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"runtime"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
+)
+
+// captureStderr builds a logger with newStdLogger while stderr is a pipe, and
+// returns everything fn logs through it. The logger resolves stderr when it is
+// built, so it has to be built inside the redirect.
+func captureStderr(t *testing.T, fn func(lg *zap.Logger)) string {
+	t.Helper()
+
+	r, w, err := os.Pipe()
+	require.NoError(t, err)
+
+	orig := os.Stderr
+	os.Stderr = w
+	defer func() { os.Stderr = orig }()
+
+	lg, _ := newStdLogger()
+	fn(lg)
+	require.NoError(t, w.Close())
+
+	out, err := io.ReadAll(r)
+	require.NoError(t, err)
+
+	return string(out)
+}
+
+func TestNewStdLogger(t *testing.T) {
+	// The configuration is read before the log settings inside it can be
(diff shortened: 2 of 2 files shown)


lentitude2tk (approved):
/LGTM

lentitude2tk (approved):
/LGTM

PR #1112, opened 2026-07-25
log: drop the unused rate limiter and use atomic.Pointer

huanghaoyuanhhy:
## Summary

`RatedDebug`, `RatedInfo` and `RatedWarn` arrived with the log package when it was copied from Milvus in 117d95f, and nothing in this repository has ever called them. They are the only readers of `R()`, which is the only reader of the rate limiter global, so the whole chain is unreachable. All of it is exported, which is why the `unused` check never had anything to say about it.

With that gone, the three globals that remain all hold pointers, so `atomic.Value` bought nothing `atomic.Pointer[T]` does not, and cost a type assertion at every read.

## Changes

- Delete `RatedDebug`, `RatedInfo`, `RatedWarn`, `R()` and the `_globalR` it reads, along with the `jaeger-client-go/utils` import they needed.
- Declare the remaining three globals as `atomic.Pointer[T]` and drop the five type assertions their readers needed, in `L()`, `S()` and the two level accessors in `global.go`.
- In `init`, call `Sugar()` on the logger already in hand rather than storing it and loading it back to reach the same value.

## Notes

`jaeger-client-go` moves from a direct requirement to an indirect one rather than leaving `go.mod`. It stays in the build graph because `milvus/pkg/v2/log` imports the same package, reached from here through `core/restore/secondary`, so no packages stop compiling — this drops our own dependency on it, not the dependency.

One semantic difference worth naming rather than leaving for someone to find: `atomic.Value` panics on a nil `Store`, `atomic.Pointer` accepts it. A nil logger would now fail at first use instead of at the store that caused it. Nothing stores nil today — the only writer is `ReplaceGlobals`, and the one caller that can fail, `InitLogger`, panics on the error before reaching it.

## Verification

`go build ./...`, `go test --race ./...` and `golangci-lint run ./...` pass, and the built binary still resolves and prints a configuration. The deleted names were checked for callers across the module including tests; outside their own definitions the only remaining mentions were their own doc comments.

/kind improvement

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -26,7 +26,6 @@ require (
 	github.com/swaggo/gin-swagger v1.6.1
 	github.com/swaggo/swag v1.16.6
 	github.com/tencentcloud/tencentcloud-sdk-go/tencentcloud/common v1.3.133
-	github.com/uber/jaeger-client-go v2.30.0+incompatible
 	github.com/vbauerster/mpb/v8 v8.12.1
 	go.etcd.io/etcd/client/v3 v3.7.0
 	go.uber.org/atomic v1.11.0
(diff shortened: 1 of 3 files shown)


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

PR #1113, opened 2026-07-26
test: drop schema version consistency gate before add_collection_field

huanghaoyuanhhy:
## Summary

Reopens #1078, which was closed when master was dropped from CI in #1080. Bringing a master-latest scenario back for secondary restore (#1077) needs this first, or the job dies in "Prepare data on upstream".

The CI helper that polls `schema_version_consistent_segments` before calling `add_collection_field` is both unnecessary and unsatisfiable on current `master-latest` Milvus images. Release images (`2.6-latest`) pass only because the stat does not exist there, so the poll short-circuits.

**It is unnecessary.** Milvus removed the proxy-side schema version consistency gate for schema-change DDLs (milvus-io/milvus#49733). Added fields are required to be nullable, and segcore fills them on older sealed segments with a virtual default/NULL column at load time, so schema-version skew across segments is tolerated by design.

**It is unsatisfiable.** The always-on backfill compaction policy was replaced by bump_schema_version compaction (milvus-io/milvus#48808), which still defaults to disabled on master. In a small CI collection that never reaches the trigger threshold for mix or sort compaction, nothing advances a sealed segment's schema version, so `schema_version_consistent_segments` stays at `0/1` and the 60s poll always times out:

```
TimeoutError: schema version is not consistent for hello_milvus_added: 0/1 segments are consistent
```

## Changes

- Remove `wait_schema_version_consistent`, `add_collection_field_with_schema_retry`, and `is_schema_version_not_ready` from `example/prepare_data.py` and `tests/testcases/test_restore_backup.py`
- Call `client.add_collection_field` directly at both call sites

/kind improvement

### DIFF
--- a/example/prepare_data.py
+++ b/example/prepare_data.py
@@ -15,85 +15,7 @@ from pymilvus import (
     Collection,
     MilvusClient,
 )
-from pymilvus.exceptions import MilvusException
 import argparse
-
-
-SCHEMA_VERSION_CONSISTENT_KEY = "schema_version_consistent_segments"
-SCHEMA_VERSION_TOTAL_KEY = "schema_version_total_segments"
-
-
-def is_schema_version_not_ready(exc):
-    message = str(exc)
-    return (
-        "schema version" in message
-        and (
-            "not ready" in message
-            or "consistency check failed" in message
-        )
-    )
-
-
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/LGTM

lentitude2tk (approved):
/LGTM

Issue #1115, opened 2026-07-26
[Bug]: server version resolves to "latest dev build" on Zilliz Cloud, enabling every feature gate


huanghaoyuanhhy:
### Milvus-Backup Version

master (54e832d)

### Can you reproduce this issue on the latest version?

Yes, it also happens on the latest version

### Current Behavior

Backing up a Zilliz Cloud instance, the client cannot determine the server version and falls
back to treating it as the newest possible build, which turns on every version-gated feature
unconditionally:

```
[INFO] [milvus/grpc.go:395] ["connect to server"] [server="Zilliz Cloud Vector Database(Compatible with Milvus 2.6)"]
[INFO] [milvus/grpc.go:427] ["get version from RPC"] [version=unknown-20260608-7ea6e3526]
[WARN] [milvus/grpc.go:475] ["cannot parse server version as strict semver, treat as latest dev build"] [version=unknown-20260608-7ea6e3526] [error="invalid semantic version"]
```

`GetVersion` returns `unknown-20260608-7ea6e3526`, which carries only a build date and a
commit — no version. The `Connect` response on the line above does name the Milvus line
("Compatible with Milvus 2.6"), but it is not consulted for the feature check, so the client
ends up with no version at all and applies `_latestDevVersion` (99.0.0).

Every entry in `_featureTuples` is then enabled regardless of what the server implements:

1. Features gated on recent patches (`FlushAll` at >= 2.6.11, `CollectionLevelGCControl` and
   `FuncRuntimeCheck` at >= 2.6.8, `MultiL0InOneJob` at >= 2.6.5) are assumed present.
   `FlushAll` in particular selects the bulk flush strategy with no fallback, so a server
   below that patch fails the backup instead of degrading to serial flush.
2. `ReplicateMessage`, the one flag with an upper bound (>= 2.5.0, < 2.6.0), is wrongly
   **disabled** on any dev build of the 2.5 line such as `2.5-20260404-31fb3fc`, because
   99.0.0 fails that upper bound.

### Expected Behavior

When `GetVersion` carries no version, the version named in the `Connect` response should be
used, so a "Compatible with Milvus 2.6" server is gated on the 2.6 line. A build tag that
identifies only a line (`2.5-20260404-...`) should stay within that line rather than
resolving above every constraint.

### Steps To Reproduce

1. Point milvus-backup at a Zilliz Cloud instance, or at any Milvus built from a detached
   HEAD where the branch name resolves to `unknown`.
[...]

PR #1116, opened 2026-07-26
milvus: detect the server version from every string it reports

huanghaoyuanhhy:
## Summary

Feature detection used a single source, the `GetVersion` RPC, and a single parser, strict
semver. On Zilliz Cloud that source answers `unknown-20260608-7ea6e3526` — a build date and a
commit, no version — so the client fell back to `_latestDevVersion` (99.0.0) and enabled every
version-gated feature. The `Connect` response names the Milvus line in the same session
("Zilliz Cloud Vector Database(Compatible with Milvus 2.6)") but was discarded.

The feature check now considers every string the server reports about itself and understands a
version that is only known down to MAJOR.MINOR.

issue: #1115

## Changes

- Split `parseVersionForFeature` into `releaseVersion` (strict semver, unchanged logic) and
  `embeddedVersion` (recovers `MAJOR.MINOR[.PATCH]` from a dev build tag or a vendor product
  description). Matching runs in two passes over the candidates, so an exact release version
  always wins over an embedded one no matter which call reported it.
- Pass the `Connect` build tags alongside the `GetVersion` result, so the source that does
  know the version decides.
- Add `_headOfLinePatch`: a version known only to MAJOR.MINOR resolves to the head of that
  line instead of jumping above every constraint. This also fixes `ReplicateMessage`
  (`>= 2.5.0-0, < 2.6.0-0`) being wrongly disabled on 2.5 dev builds such as
  `2.5-20260404-31fb3fc`, which previously resolved to 99.0.0 and failed the upper bound.
- `_latestDevVersion` now applies only when no candidate contains a version at all, e.g.
  `master-20260226-abcdef` or `unknown-20260608-7ea6e3526`.
- The version regex is anchored to the start of the string or to a `Milvus` prefix, so a
  number elsewhere in a product name is not read as the server version.

### DIFF
--- a/internal/client/milvus/grpc.go
+++ b/internal/client/milvus/grpc.go
@@ -10,6 +10,7 @@ import (
 	"math"
 	"net"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/LGTM

Issue #1107, opened 2026-07-25
Notice: deprecations and migrations for the next release


huanghaoyuanhhy:
**Status: shipped in v0.6.0 (2026-09-16).** Everything below is live in v0.6.0; the release notes cover the same ground.

This is the single place where deprecations and migrations for that release are collected. It is updated as further ones land, so watch this issue rather than re-reading the diff. Every entry answers the same three questions: does the old form still work, what replaces it, and what do I actually run.

## Index

| Change | Old form still works? | Entry |
| --- | --- | --- |
| The configuration file moves to schema v2 | Yes — translated while loading, with a warning | [Configuration schema v2](#configuration-schema-v2) |
| The CLI flags deprecated in 0.5 are rejected | No — the command stops and names the replacement | [Deprecated CLI flags are rejected](#deprecated-cli-flags-are-rejected) |

## Configuration schema v2

The configuration file is now versioned, and v2 is the schema the program runs on. It answers the configuration complaints collected in #637: the file was long and hard to fill in correctly, and it was organized around MinIO no matter what you actually ran. The design discussion and the complete key-by-key mapping are in #1089; this entry is the operator's version of it.

### Your existing file keeps working

A v1 file is still accepted. It is translated to v2 while loading, and the program says so once per run:

```
cfg: loaded a v1 configuration and translated it to v2; run `milvus-backup config migrate` to convert the file itself
```

Nothing is required of you for the file to keep loading. When you want to stop relying on that translation, convert it:

```bash
milvus-backup config migrate --config backup.yaml --output backup-v2.yaml
```

The result is a complete v2 file. Add `--strict` to fail rather than write a file that does not validate.

### Environment variables do not carry over, and they fail silently

This is the part worth reading twice.

Environment variables belong to the schema version the file is written in. A file declaring `configVersion: v2` is resolved with v2 names only, and a v1 variable set against a v2 file is ignored without a warning:

```console
$ MINIO_BUCKET_NAME=my-bucket milvus-backup config show --config backup-v2.yaml
Milvus.Storage.BucketName    a-bucket    config    milvus.storage.bucketname
```

The file wins because `MINIO_BUCKET_NAME` is not a name the v2 schema knows. If you inject bucket names or credentials through the environment — Helm values, docker-compose, a CI job — convert those in the same change as the file, not afterwards. Otherwise a backup can quietly run against the wrong bucket or with the wrong credentials.

Two commands make this checkable:

[...]

Issue #617, opened 2025-04-14
[Bug]: Milvus proxy address effect is abnormal


maxhoov:
### Current Behavior

Milvus: 2.5.7
milvus-backup: 0.5.3

backup.yaml:
```
......

http:
  simpleResponse: true

milvus:
  address: milvus.proxy
  port: 19530
  user: "admin"
  password: "******"
  tlsMode: 0
  caCertPath: ""
  serverName: ""
  mtlsCertPath: ""
  mtlsKeyPath: ""

......

backup:
  maxSegmentGroupSize: 2G

  parallelism: 
    backupCollection: 4
    copydata: 128
    restoreCollection: 2
  
  keepTempFiles: false
  
  gcPause:
    enable: true
    seconds: 7200
    address: http://localhost:9091
```
When creating a new backup(milvus-backup create -n ...), the milvus proxy address defined in backup.yaml will be automatically appended with an inexplicable address (tcp://169.169.215.215:9530), which is not used anywhere in the system.

Full logs:
```
/app #./milvus-backup create -n milvus20250410
v0.5.3(Built on 2025-04-02T05:31:21Z from Git SHA f20e3ba)
config:backup.yaml
[2025/04/1014:17:43.433+00:00][INFO][logutil/logutil.go:165]["Log directory"][configDir-]
[2025/04/1014:17:43.433+00:00][INFO][loguti1/loguti1.go:166][Set log file to“][path-logs/backup.log]
[2025/04/1014:17:43.434+00:00][INPO][core/backup impl create backup.go:44]["receive CreateBackupRecuest"][reguest
1d-917b2c20-1616-11f0-82b4-32a2624809de][backupName=milvus_20250410][collections-"[]"][databaseCollections=][async=false][force=false][metaOnly=false]
[2025/04/1014:17:43.434+00:00][INFO][core/backup_context.go:87]["{Base:0xc000277508 MaxSegmentGroupSize:2147483648
BackupcollectionParallelism:4 BackupCopyDataParallelism:128 RestoreParallelism:2 KeepTempFiles:false GcPauseEnable:false GcPauseSeconds:7200 GcPauseAddress:http://localhost:9091}"]
[2025/04/1014:17:43.434+00:00][INFO][core/backup_context.go:88]["{Base:0xc000277508 Enabled:true DebugMode:false S
impleResponse:true}"]
[...]

Issue #197, opened 2023-09-08
[Bug]: run milvus-backup in k8s pod, retrurn "failed to connect \"Addr\": \"milvus:tcp:\" "


jackiedlh:
### Current Behavior

When I run milvus-backup command on k8s pod, the target milvus address is wrong.

in backup.yaml:

milvus:
  address: milvus
  port: 19530
  authorizationEnabled: true
  # tls mode values [0, 1, 2]
  # 0 is close, 1 is one-way authentication, 2 is two-way authentication.
  tlsMode: 0
  user: ""
  password: ""

in log:

[2023/09/08 07:35:33.943 +00:00] [INFO] [logutil/logutil.go:165] ["Log directory"] [configDir=]
[2023/09/08 07:35:33.943 +00:00] [INFO] [logutil/logutil.go:166] ["Set log file to "] [path=logs/backup.log]
[2023/09/08 07:35:33.943 +00:00] [INFO] [core/backup_impl_create_backup.go:25] ["receive CreateBackupRequest"] [requestId=4b980b8b-4e1a-11ee-8398-82b39bdfe663] [backupName=backup] [collections="[]"] [databaseCollections=] [async=false]
[2023/09/08 07:35:34.168 +00:00] [WARN] [grpclog/grpclog.go:46] ["[core][Channel #1 SubChannel #2] grpc: addrConn.createTransport failed to connect to {\n  \"Addr\": \"milvus:tcp:\",\n  \"ServerName\": \"milvus:tcp:\",\n  \"Attributes\": null,\n  \"BalancerAttributes\": null,\n  \"Type\": 0,\n  \"Metadata\": null\n}. Err: connection error: desc = \"transport: Error while dialing dial tcp: address milvus:tcp:: too many colons in address\""]

### Expected Behavior

expect connect to:

milvus:19530

### Steps To Reproduce

```markdown
1 copy milvus-backup tool to k8s milvus proxy pod
2 copy config files to milvus pod
3 run command:  ./milvus-backup create -n backup
```

### Environment

```markdown
aws EKS
```

wayblink:
address should be a ip or a domain.

Archalbc:
> address should be a ip or a domain.

"milvus" is correct in a Kubernetes context, it will resolve to the milvus-proxy pod, actually this look like a bug.

nightmorning0:
It is definitely a bug. Add PORT# at the tail of the address. Some thing like this:
```
milvus:
address: milvus:19530#
[...]

PR #1118, opened 2026-07-27
cfg: keep v2 environment variables for credentials only

huanghaoyuanhhy:
## Summary

v2 gave every parameter an environment variable — 63 names for 64 config keys, where v1 had 40. Most of them are a liability rather than a feature.

Kubernetes injects a variable for every Service in the namespace, so a Service named `milvus` becomes `MILVUS_PORT=tcp://10.0.0.1:19530` in every pod. That is exactly the name a connection parameter claims, and an environment variable outranks the config file, so the injected value silently replaces the configured one. v1 named that parameter `MILVUS_PORT` and the collision was reported twice as a connection failure — #197 and #617, both closed without addressing the cause.

A credential is different: it has to come from outside the file (a Kubernetes Secret, a CI secret store, an operator), and credential names are not at risk, because the injected names always end in `_PORT`, `_HOST` or `_SERVICE_*`. So v2 keeps an environment variable for credentials only — 15 names — and everything else is named by its config key, overridable per run with `--set milvus.grpc.port=29530`.

Only the v2 schema changes. A v1 file is still resolved with the v1 schema, v1 environment variables included, so no released deployment loses anything: v2 landed after v0.5.16 and has never shipped.

Related to #1107

## Changes
- drop `EnvKeys` from every v2 parameter that is not a credential, and record the reason in the package documentation
- point the v1 environment variable migration table at the config key that replaced each one, and word the migration report as a replacement rather than a rename
- rewrite `docs/user_guide/env_variables.md` around the credentials-only rule, including why
- cover the rule in the loader tests: a connection variable is inert, a credential variable applies

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,10 +67,11 @@ Use values that match the Milvus deployment. In common installations, the storag
 | `milvus.storage.bucketName` | `a-bucket` | `milvus-bucket` |
 | `milvus.storage.rootPath` | `files` | `file` |
 
-Configuration values can also be supplied through [environment variables](docs/user_guide/env_variables.md) or overridden with `--set`:
+Credentials can also be supplied through [environment variables](docs/user_guide/env_variables.md), and any value can be overridden for a single run with `--set`:
 
 ```shell
 milvus-backup --set MILVUS_USER=root --set MILVUS_PASSWORD=Milvus list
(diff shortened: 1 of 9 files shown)


lentitude2tk (approved):
/LGTM

Issue #1119, opened 2026-07-27
[Enhancement]: use snapshot export/restore as the backup format on Milvus 3.0+


huanghaoyuanhhy:
## What would you like to be added?

A second backup format for Milvus 3.0+, built on the snapshot export/restore APIs, selected by a hard version fork rather than layered onto the existing one.

**Milvus < 3.0** keeps the current path unchanged: flush, `GetPersistentSegmentInfo`, path reconstruction, copy into the `backuppb` archive, restore by bulk-insert. Nothing about it changes.

**Milvus >= 3.0** takes a different path entirely, where milvus-backup becomes an orchestrator and Milvus moves the bytes:

```
backup:   flush → CreateSnapshot(db, coll, name)
                → ExportSnapshot(db, coll, name, target_s3_path=<backup bucket>) → job_id
                → poll GetExportSnapshotState → snapshot_metadata_uri
                → record the URI in the backup meta
                → DropSnapshot
                → RBAC and multi-collection orchestration as today

restore:  read the meta → per-collection snapshot_metadata_uri
                → RestoreExternalSnapshot(uri, target_db, target_collection) → job_id
                → poll GetRestoreSnapshotState
                → RBAC restore as today
```

The two formats never have to interoperate, because restoring a backup onto an older Milvus than it came from is already outside what this tool supports. That constraint is currently a convention rather than something the code enforces, so the fork should also add an explicit version gate with a clear error, instead of letting a 3.0-format backup fail somewhere deep in the restore path.

## Why is this needed?

**The copy window becomes a server-enforced guarantee instead of a best-effort advisory.** `PauseGC` (`core/backup/gc_controller.go`) asks the whole cluster to pause GC for a fixed hour, logs a warning and continues if that fails, does not stop compaction, and does not stop `DropCollection`. A snapshot holds its referenced segments against GC unconditionally for as long as it exists, and `ExportSnapshot` pins the snapshot for the duration of the copy so nothing can drop it mid-export.

**Index files are carried, so restore does not rebuild indexes.** `ListSnapshotDataFiles` collects insert, stats, delta and bm25 binlogs, vector and scalar index files, text and JSON key index files, and StorageV2/V3 manifest objects. milvus-backup copies only insert and delta binlogs today and leaves indexes to be rebuilt on restore, which for a large collection is usually the dominant cost. The trade is a larger archive — index files are not small — but it is the right default for a backup.

[...]

PR #1120, opened 2026-07-27
client: add snapshot rpcs to the milvus grpc client

huanghaoyuanhhy:
## Summary

Part 1 of the snapshot-based collection backup work. Adds the client surface for Milvus snapshots — proto bump, feature flag, and five RPC wrappers. Plumbing only: nothing calls these yet and no existing behavior changes.

The backup path will use them to freeze a collection with a server-enforced guarantee, replacing the best-effort `PauseGC` advisory it relies on today.

## Changes

- Bump `milvus-proto/go-api/v2` to a master-line build. The `v2.6.x` tags are cut from a branch that carries no snapshot code, and the previously pinned build predates `PinSnapshotData` / `UnpinSnapshotData`. Field numbers match the v3 module Milvus master now uses, so the wire format is unchanged.
- Add a `Snapshot` feature flag constrained to `>= 3.0.0-0`. Snapshot landed on the 3.0 line and the 2.6 branch carries none of it, so this is a clean major-line boundary rather than a patch one. Every snapshot call fails with one recognizable error when the flag is off, which keeps an older server on the existing path instead of hitting an opaque unknown-method from gRPC.
- Wrap `CreateSnapshot`, `DropSnapshot`, `DescribeSnapshot`, `PinSnapshotData` and `UnpinSnapshotData` on the `Grpc` client.
- `PinSnapshotData` rejects a TTL below one second. The wire field is in seconds and reads 0 as "never expires", and no API can release such a pin if the caller dies before unpinning, so a duration that would round down to zero is refused in the client rather than silently sent.

`DescribeSnapshot` is included because its `S3Location` is what locates the snapshot metadata file, which is where the segment list lives. The server writes that path during the normal create, not only for exported snapshots, so the later manifest reader needs neither the collection id nor the snapshot id to find it.

Related to #1119


lentitude2tk (approved):
/LGTM

PR #1121, opened 2026-07-27
ci: close issues stuck waiting on the reporter

huanghaoyuanhhy:
## Summary

Adds a stale workflow that closes issues which have been waiting on the reporter, plus a companion workflow that clears the label as soon as they answer, plus a doc describing the resulting state machine.

The scope is deliberately narrow. `only-labels` means an issue is touched only while it carries `triage/needs-information` — design proposals, feature requests and the rest of the backlog are never seen by this workflow no matter how long they sit, and pull requests are excluded outright so long-lived dependabot PRs are left alone.

`triage/needs-information` is treated as a state bit meaning "waiting on the reporter", not as a one-way marker. Once the reporter answers, the ball is back with us and the label has to come off — otherwise the countdown keeps running and eventually closes an issue that is actually waiting on a maintainer. Doing that by hand works right up until the day somebody forgets, so `clear-needs-information.yaml` does it instead.

The label is cleared when the person who opened the issue comments, or when anyone outside the org comments. Both conditions are needed: checking only `author_association` would miss an org member replying on their own issue, and checking only the issue author would miss a third party supplying the missing details. A maintainer commenting deliberately does not clear it — asking a follow-up question means we are still waiting — though it does reset the 30 day clock, which is the intended behaviour.

Timings are 30 days to mark stale and a further 14 to close, so 44 days from the last activity. Issues are closed as `not_planned` with a message stating that this is not a judgement about whether the problem is real, and inviting a reopen with the requested details.

`docs/issue_lifecycle.md` writes the whole thing down: what the label means, the timings, which comments reset which clock, what is explicitly out of scope, what to do if your issue was closed this way, and when maintainers should apply the label. Both bot messages link to it, since the people most affected arrive from a GitHub notification rather than by browsing the repository.

## Changes

- add `.github/workflows/stale.yaml`, scoped to `triage/needs-information` via `only-labels`, with pull requests excluded through `days-before-pr-stale/close: -1`
- add `.github/workflows/clear-needs-information.yaml`, removing `triage/needs-information` and `stale` when the reporter or an outside contributor comments
- add `docs/issue_lifecycle.md` documenting the state machine, and link it from the stale and close comments
- set `remove-stale-when-updated` explicitly rather than relying on the default, since the whole reset behaviour depends on it
- expose a `dry-run` input on `workflow_dispatch` so the first run can be verified with `debug-only` before anything is actually closed

/kind improvement

### DIFF
--- a/.github/workflows/clear-needs-information.yaml
+++ b/.github/workflows/clear-needs-information.yaml
@@ -1,0 +1,52 @@
+name: Clear needs-information
+
+# `triage/needs-information` is a state bit meaning "waiting on the reporter".
+# Once they answer, the ball is back with us and the label has to come off —
+# otherwise stale.yaml keeps counting down and eventually closes an issue that
+# is actually waiting on a maintainer. Removing it by hand works right up until
+# the day somebody forgets, so it is done here instead.
+#
+# The label is cleared when either:
+#   - the person who opened the issue comments (they are who we were waiting on), or
+#   - anyone outside the org comments (a third party supplying details).
+#
+# A maintainer commenting does not clear it: asking a follow-up question means we
+# are still waiting. `stale` is removed at the same time so the countdown does not
+# linger until the next scheduled run.
+
+on:
+  issue_comment:
+    types: [created]
+
+permissions:
+  issues: write
+
+jobs:
+  clear:
+    if: >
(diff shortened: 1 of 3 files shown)


lentitude2tk (approved):
/LGTM

PR #1123, opened 2026-07-27
backup: skip external collections

### LINKED ISSUE #1064: [Bug] Full restore should skip external collections instead of blocking internal collection restore
> *This was generated by AI during triage.*

## Summary

`milvus-backup restore` should skip external collections/tables during full restore. External collections are not physically restorable by design, but the current full restore attempts to restore them and can block internal collections in the same backup from being restored.

## Current Behavior

- Full backup of an instance containing both internal and external collections succeeds.
- Full restore attempts to recreate the external collection schema.
- Restore fails on the external collection because the backup contains the internal virtual PK field:

```text
field name "__virtual_pk__" is reserved for internal use and cannot be used in user schemas: field name invalid[field=__virtual_pk__]
```

- If the external collection is restored before an internal collection, the restore context is canceled and the internal collection restore is aborted.
- If the internal collection happens to restore first, it can be restored, but the full restore command still fails on the external collection.
- Filtering the restore to only the internal collection succeeds, confirming the internal collection backup itself is usable.

## Expected Behavior

Since physical backup/restore for external collections is unsupported by design, full backup/restore should explicitly skip external collections/tables, or at least isolate their failure so unsupported external collections do not block internal collections from restoring.

## Steps to Reproduce

1. Deploy Milvus with external table support.
2. Create one normal/internal collection with data and vector index.
3. Create one external collection/table backed by external files.
4. Run a full backup with `milvus-backup create`.
5. Drop both collections.
6. Run a full restore with `milvus-backup restore`.
7. Observe that restore fails on the external collection; when the external collection is first in the restore plan, the internal collection restore is canceled.

## Reproduction Details

- Milvus image: `harbor.milvus.io/manta/milvus:master-20260611-35079b9`
- Milvus git commit: `35079b98df`
- Local milvus-backup build: `v0.5.9-161-gfd47b01`
- Also reproduced with latest release: `v0.5.16`
- Internal-only filtered restore result from the same backup: `count(*) = 1000`, vector search `topK=5` returned 5 hits.

## Impact

Full backups are not reliably restorable for mixed instances containing both internal and external collections. A user expecting internal collections to be recoverable from a full backup may get a failed restore because of an unsupported external collection in the same backup.

huanghaoyuanhhy:
## Summary

An external collection only maps to data owned by an external source, so Milvus holds no binlog for it. Backing one up produces an entry that can never be restored — restore fails on the reserved `__virtual_pk__` field and, when the external collection comes first in the restore plan, takes the internal collections in the same backup down with it.

Skip external collections at backup time so they never enter the backup in the first place.

## Changes

- Add `Task.excludeExternalColl`, which describes every listed collection and drops the external ones before the database/collection tasks run, logging the skipped namespaces.
- Detect external collections with `typeutil.IsExternalCollection` from `milvus/pkg/v2` instead of reimplementing the check.

## Notes

This covers the backup side. Backups taken before this change may still contain an external collection; restoring one of those needs a collection filter that excludes it.

Fixes #1064


lentitude2tk (approved):
/LGTM

Issue #1127, opened 2026-07-28
[Bug]: l0compact fails on Milvus 2.5 backups with "unmarshal extras"


huanghaoyuanhhy:
### Milvus-Backup Version

main (81d230f)

### Can you reproduce this issue on the latest version?

Yes, it also happens on the latest version

### Current Behavior

`milvus-backup l0compact` fails against every backup taken from a Milvus 2.5 instance. It copies
the binlogs to the output backup, then aborts and deletes them again:

```
Error: l0compact: unmarshal extras: json: cannot unmarshal bool into Go value of type string
```

The command exits 1 and prints its usage text, so the output backup is left empty.

This is not intermittent — the nightly `test-backup-restore-api (2.5, L2)` job has failed on
every run since 2026-07-17, with all 8 `testcases/test_l0compact.py` cases hitting the same
error. The 2.6 matrix is unaffected and passes.

### Expected Behavior

`l0compact` folds the L0 segments into per-segment deltalogs and exits 0, the same as it does
on Milvus 2.6.

### Steps To Reproduce

```
1. Deploy Milvus 2.5 (2.5-latest).
2. Create a collection, insert rows, delete some of them so L0 segments exist.
3. milvus-backup create -n mybackup
4. milvus-backup l0compact -n mybackup -o mybackup_l0c
5. See the error above.
```

Equivalently: run `pytest --tags L2 testcases/test_l0compact.py` against a 2.5 instance.

### Environment

```
Milvus: 2.5-latest (docker-compose standalone)
milvus-backup: main
OS: ubuntu-latest (GitHub Actions runner)
```

### Anything else?

The descriptor extras of a v1 insert binlog are not a string→string map. Milvus writes
`nullable` as a bool alongside string entries such as `original_size`:

```go
// milvus internal/storage/binlog_writer.go, NewInsertBinlogWriter
// store nullable in extra for compatible
descriptorEvent.AddExtra(nullableKey, nullable) // nullable is a bool
```

so the extras JSON looks like `{"original_size":"12345","nullable":true}`, while
`internal/l0compact/envelope.go` decodes it into `map[string]string`.

Milvus 2.6 is unaffected because its insert logs use storage v2 (bare parquet), which
`ReadInsertPK` handles through the `KindV2` branch without going through the v1 envelope
parser at all. Only the 2.5 (`KindV1`) path reaches the failing decode.

Failing nightly run: https://github.com/zilliztech/milvus-backup/actions/runs/30233909370

PR #1128, opened 2026-07-28
l0compact: decode descriptor extras as a heterogeneous map

huanghaoyuanhhy:
## Summary

`l0compact` fails on every backup taken from Milvus 2.5 with:

```
Error: l0compact: unmarshal extras: json: cannot unmarshal bool into Go value of type string
```

The descriptor extras of a v1 binlog are not a string-to-string map. Milvus writes `nullable`
as a bool next to string entries such as `original_size`:

```go
// milvus internal/storage/binlog_writer.go, NewInsertBinlogWriter
// store nullable in extra for compatible
descriptorEvent.AddExtra(nullableKey, nullable) // bool
```

so the JSON is `{"original_size":"12345","nullable":true}`, while `envelope.go` decoded it into
`map[string]string`. `ReadInsertPK` reaches that decode for every `KindV1` segment, so the
command is unusable against 2.5. Milvus 2.6 stores insert logs as bare parquet and takes the
`KindV2` branch, never touching the v1 envelope parser — which is why only the 2.5 matrix
failed.

Typing the field as `map[string]any` is what the wire format actually carries. The sole
consumer compares `Extras["version"]` against `MULTI_FIELD`; comparing an `any` to a string is
still an equality on the dynamic value, so that keeps working.

This has broken the nightly `test-backup-restore-api (2.5, L2)` job on every run since
2026-07-17 — all 8 `test_l0compact.py` cases, never green since the feature landed.

## Changes

- envelope: decode descriptor extras as `map[string]any`, with a note on why they are mixed
- envelope: `BuildV1Envelope` takes the same type, keeping the writer symmetric with the reader
- test: cover extras carrying a bool, and a `version` marker sitting next to one

Both new tests fail on `main` with the exact error above and pass with the fix.

issue: #1127

### DIFF
--- a/internal/l0compact/deltalog.go
+++ b/internal/l0compact/deltalog.go
@@ -117,7 +117,7 @@ func WriteDeltalog(entries []DeleteEntry, kind StorageKind, t PKType) ([]byte, e
 			return nil, err
 		}
 		// json variant: PayloadDataType = String(20), field id 0, no version extra.
-		return BuildV1Envelope(eventDelete, 20, 0, map[string]string{}, payload)
+		return BuildV1Envelope(eventDelete, 20, 0, map[string]any{}, payload)
 	default:
 		return nil, fmt.Errorf("l0compact: unknown storage kind %d", kind)
 	}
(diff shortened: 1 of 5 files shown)


lentitude2tk (approved):
/LGTM

PR #1129, opened 2026-07-28
deps: migrate to milvus go-api v3 and pkg v3

huanghaoyuanhhy:
## Summary

Moves the module off `milvus-proto/go-api/v2` and `milvus/pkg/v2` onto the v3 lines.

`ExportSnapshot` and `RestoreExternalSnapshot` — the RPCs the 3.0 backup path is being built on — exist only in `go-api/v3`. The v2 line stopped receiving updates once Milvus master moved to v3, and no `v2.6.x` tag carries them at all, so staying on v2 caps what can be built against a 3.0 server.

The change is mechanical. Import paths only, across 48 files; not a single call site needed adjusting, which is what you would expect from a module-path major bump rather than an API break. `go build`, `go vet` and the full unit suite pass unchanged.

## Changes

- Replace `milvus-proto/go-api/v2` with `.../v3` and `milvus-io/milvus/pkg/v2` with `.../pkg/v3` throughout, including the `.mockery.yaml` source package.
- Regenerate `milvusServiceClient_mock.go` and `grpc_mock.go`. v3 declares RPCs the v2 service interface did not (`AddCollectionStructField` among others), so the old mock no longer satisfies the interface.
- Bump the `go` directive from `1.26` to `1.26.5`, which `pkg/v3` requires. CI (`go-version: 1.26`) and the Dockerfile (`golang:1.26`) both track the 1.26 line rather than a pinned patch, so neither needs a change.

## Risk

Wire compatibility with pre-3.0 servers is the thing worth watching: the client now speaks v3-generated protobufs to servers that may be anywhere from 2.2 to 2.6. The message definitions are shared between the two module lines, so the encoding should be identical, but the cross-version jobs in CI cover exactly that matrix and are the real check.

Related to #1119


lentitude2tk (approved):
/LGTM

PR #1130, opened 2026-07-28
test: regenerate the stale mocks with the current mockery

huanghaoyuanhhy:
## Summary

Three mock files were generated by an older mockery than the rest of the tree. Every `make gen-mock` run rewrote their expecter helpers from `interface{}` to `any`, so any PR that regenerated mocks for its own reasons picked up unrelated churn in these files and had to revert them by hand.

Regenerating them is a no-op for behaviour — `interface{}` and `any` are the same type — but it means the next PR that touches a mocked interface shows only its own diff.

## Changes

- Regenerate `core/restore/collMapper_mock.go`, `internal/client/milvus/manage_mock.go` and `internal/storage/client_mock.go` with mockery v3.7.1, the version the rest of the generated mocks already use.

`go build`, `go vet` and the full unit suite pass unchanged.

/kind improvement

### DIFF
--- a/core/restore/collMapper_mock.go
+++ b/core/restore/collMapper_mock.go
@@ -62,7 +62,7 @@ type MockCollMapper_TagetNS_Call struct {
 
 // TagetNS is a helper method to define mock.On call
 //   - ns namespace.NS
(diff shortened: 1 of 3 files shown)


lentitude2tk (approved):
/LGTM

PR #1131, opened 2026-07-29
client: add snapshot export rpcs and drop the unused pin rpcs

huanghaoyuanhhy:
## Summary

Adds the four snapshot RPCs the Milvus 3.0 backup path needs to move bytes: `ExportSnapshot`, `GetExportSnapshotState`, `RestoreExternalSnapshot` and `GetRestoreSnapshotState`, and removes the two that turned out to have no caller. #1120 added the create/drop/describe half; this settles the rest of the client surface. Plumbing only — nothing calls these yet, and everything stays behind the existing `Snapshot` feature flag, so a pre-3.0 server keeps taking the current path.

Export and restore are both asynchronous: the RPC durably accepts a job and returns its id, and the metadata URI that restore needs is published through the job state once the export completes. So each side is a submit plus a poll rather than one blocking call.

`PinSnapshotData` / `UnpinSnapshotData` go away. #1120 added them in case the backup path had to hold a snapshot open across the copy, and it does not: DataCoord pins the source snapshot inside `ExportSnapshot`'s submit path — before the job record is persisted — and releases it only when the job is terminal, so the queue wait is covered as well, which matters given `exportMaxConcurrentJobs` defaults to 1. A pin is not what keeps files from being GC'd either; that is inherent to the snapshot existing. It only additionally survives `DropSnapshot` and the cascade behind `DropCollection`. Create a snapshot immediately before submitting its export and there is no window left for a caller-held pin to cover.

The server half is milvus-io/milvus#50978, which is still open — nothing here runs end to end until it lands. The proto half does not depend on it: the async export job APIs are already on `milvus-proto` master (milvus-io/milvus-proto#644), which is what this bumps `go-api/v3` for. Field 2 of `ExportSnapshotResponse` (`snapshot_metadata_uri`) is deprecated and unset; job id is field 3.

Related to #1119

## Changes

- bump `milvus-proto/go-api/v3` to `v3.0.0-20260729035609-9f7d27f3eb7f` for the async export job APIs
- add `ExportSnapshot` / `GetExportSnapshotState` / `RestoreExternalSnapshot` / `GetRestoreSnapshotState` to the gRPC client, with `ExportSnapshotInput` and `RestoreExternalSnapshotInput` carrying the target path and the optional external storage spec
- a job that failed is returned through the info, not as an error — the state query succeeded, and reading `state` and `reason` is how the caller learns the job failed, so flattening it would leave no way to tell the two apart
- both calls that take a db send it through the `dbname` header like the rest of this client rather than the request field; Milvus fills that header into both request types in its database interceptor
- drop `PinSnapshotData` and `UnpinSnapshotData`, which have no caller and no window left to cover
- regenerate the affected mocks

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -17,7 +17,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware v1.4.0
 	github.com/huaweicloud/huaweicloud-sdk-go-v3 v0.1.205
 	github.com/imroc/req/v3 v3.59.0
-	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260727091820-f95888a4df32
+	github.com/milvus-io/milvus-proto/go-api/v3 v3.0.0-20260729035609-9f7d27f3eb7f
 	github.com/milvus-io/milvus/pkg/v3 v3.0.0-20260728030618-14564e26159c
 	github.com/minio/minio-go/v7 v7.2.1
 	github.com/samber/lo v1.53.0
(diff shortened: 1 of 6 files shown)


lentitude2tk (approved):
/LGTM

PR #1132, opened 2026-07-30
backup: add the format field and the snapshot meta layout

huanghaoyuanhhy:
## Summary

Groundwork for the Milvus 3.0 backup path (#1119), where a collection's data is an exported snapshot bundle rather than binlogs copied file by file. Nothing writes this format yet — this only teaches the meta layer what it looks like, and stops restore from misreading one.

`BackupInfo.format` records which format a backup was taken in, and `CollectionBackupInfo.snapshot_backup` says where a collection's bundle is.

**`format` is a string, not an enum.** This meta is stored as plain json through `encoding/json`, where an enum lands in the file as a bare integer that cannot be read without the `.proto`. `load_state` is a string for the same reason. Empty is the binlog format — what every backup written before this field reads back as — so nothing has to be migrated.

**`metadata_path` is relative to the backup directory**, not the absolute URI the export job returns. A backup directory is expected to be movable, and an absolute URI would pin it to the bucket and prefix it was written to. Milvus supports the move: for a self-contained bundle it derives the root from the metadata URI the caller passes and rebases the stored paths, so restore can rebuild the URI from whatever backup storage it is running against.

**A snapshot-format backup stops after `collection_meta.json`.** It writes `full_meta.json`, `backup_meta.json` and `collection_meta.json`, and not the partition and segment files — it has no segments to put in them. Writing those two empty would give the backup a second read path through `levelToTree`, which derives a collection's size by summing its segments and would report zero for a format whose size comes from the export job. `backup_meta.json` stays the one marker that says a directory holds a backup, in either format, so `Exist()` is untouched.

An older milvus-backup can still read such a backup — it would find no segments and report a successful restore of an empty collection — and nothing in the meta can prevent that. So the guard is on this side: `restore.NewTask` now refuses a format the binlog path cannot restore, and both the CLI and the HTTP entry point go through it.

Related to #1119

## Changes

- add `BackupInfo.format`, `CollectionBackupInfo.snapshot_backup` and `SnapshotBackupInfo` to the backup proto, plus `format` on `BackupInfoBrief` so `list` and `get` show it
- stop `meta.Write` after `collection_meta.json` for the snapshot format, and carry `format` in the `backup_meta.json` view so a reader that only fetches it knows what it has
- refuse an unrestorable format in `restore.NewTask`
- assert in the tests that the partition and segment files stay unwritten, and that the stored json says `"format":"snapshot"` rather than an integer

## Notes for review

The `.pb.go` diff is ~800 lines because inserting a message reshuffles the descriptor blob and the internal indices; the actual additions are the three fields and one message. The generator was run once with no proto changes first, to confirm the toolchain reproduces the committed file byte for byte before anything was edited.

Regenerating the swagger also picks up three fields that were added to the proto without it: `ezk_mapping`, `RestoreCollectionMapping.override` and `RestoreCollectionOverride`. Hand-trimming generated output seemed worse than carrying them.


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

PR #1133, opened 2026-07-30
backup: add a snapshot strategy that exports bundles

huanghaoyuanhhy:
## Summary

The Milvus 3.0 half of #1119. `strategy=snapshot` freezes each collection into a snapshot, has Milvus export it into the backup directory as a self-contained bundle, and records where the bundle landed. No binlog bytes pass through milvus-backup — it orchestrates and Milvus moves the data.

Per backup: one `FlushAll`. Then per collection: `CreateSnapshot`, `DescribeSnapshot` for the snapshot's own boundary, `ExportSnapshot`, poll `GetExportSnapshotState` until the job is terminal, record the bundle, drop the snapshot.

`StrategyAuto` does not pick this yet, even on a server that reports the snapshot feature. That flag is a version constraint, and the server side landed in stages: Milvus master serves `CreateSnapshot` and `DescribeSnapshot` today but answers `ExportSnapshot` with `service unimplemented`, so routing auto there fails every backup against master — CI caught exactly that. Auto can move once the export is released and the constraint can name a version that has it.

## Why it is shaped this way

**The flush is not redundant.** A snapshot admits only the segments below its channel seek positions, so without one it would hold whatever the last automatic flush left behind rather than the collections as they are now. One `FlushAll` covers all of them, the same call the bulk flush strategy already makes.

The bound this leaves is worth stating: a collection whose snapshot is taken well after the flush is bounded by its own channel checkpoint at creation time, so writes since the flush are in it only as far as automatic flushing has carried them. How far behind the last collection runs depends on `dataCoord.snapshot.exportMaxConcurrentJobs`.

**Each collection creates its snapshot immediately before the export that pins it.** DataCoord takes the pin when it *accepts* the job, not when it starts copying, so a create-then-submit pair leaves no window where a snapshot sits unpinned waiting its turn. The collection semaphore bounds how many exports are in flight, which matters because the timeout starts at accept too: anything queued beyond what `dataCoord.snapshot.exportMaxConcurrentJobs` can run burns that budget waiting. That setting is refreshable and has no upper bound — raising it is the lever if the tail lags — but it is 1 until someone raises it.

**Dropping the snapshot retries.** DataCoord releases the export's pin on its own reconcile tick rather than inline with the state change the poll observed, so the first drop after a job completes can still be refused as pinned.

**GC is not paused on this path.** Pausing it is how the binlog path holds files still while it copies them one by one. A snapshot does that itself, unconditionally, and only for the collection being backed up.

**The bundle goes under a `bundle/` directory of the backup**, so everything Milvus writes — its metadata under `snapshots/`, its manifests, and the data files under `files/` — sits under one prefix beside the `meta/` this tool writes:

```
backup/<name>/
├── meta/{full,backup,collection}_meta.json
└── bundle/
    ├── snapshots/{collID}/metadata/{snapID}.json
    ├── snapshots/{collID}/manifests/{snapID}/{segID}.avro
    └── files/insert_log|delta_log|index_files/...
```

The layout is not baked in anywhere. What the meta records is the metadata path *relative to the backup directory*, so each backup describes where its own bundle is; restore joins the two and never has to know the convention. Moving it later costs one line and leaves existing backups restorable.

**The export target is built from the backup storage config**, in one of the two shapes Milvus parses:

| configured endpoint | URI | endpoint comes from |
|---|---|---|
| yes | `minio://<endpoint>/<bucket>/<key>` | the URI host, verbatim |
| no | `s3://<bucket>/<key>` | Milvus, derived from `cloud_provider` + `region` |

The configured endpoint wins whenever there is one: derivation only produces a provider's canonical public endpoint, which is wrong for a deployment reached over an internal or private-link address — and wrong in a way that surfaces as a copy failure rather than a configuration error. `s3`, `aws`, `minio`, `tencent`, `aliyun` and `hwc` are mapped onto the `cloud_provider` values Milvus accepts; anything else is refused here, where the error can name the provider.

`extfs.cloud_provider` is always sent. Milvus infers one from the URI scheme when it is missing, and its own comment calls that inference a source of silent misconfiguration since `s3://` covers both AWS and a self-hosted store. The rest of the spec is left empty when both sides are the same backend, so Milvus writes with its own credential and the access key stays off the wire.

**The cluster info the flush returns is recorded**, the same way the bulk flush strategy records it — control channel, physical channels, per-channel flush messages — so both formats describe the cluster they came from the same way. Per-collection channel checkpoints are not recorded: a snapshot's boundary is its own channel seek positions, which no API exposes, and the flush response describes where the flush left the channels instead. Anything that depends on those stays off for this format.

Related to #1119

## Changes

- add `StrategySnapshot` and `snapshotStrategy`, plus the per-collection `collSnapshotTask` that drives create → export → poll → record → drop
- build the export target URI and external spec from the backup storage config, and record the bundle's metadata path relative to the backup directory
- record `format` in the meta, and skip the partition and segment meta files for this format, matching what `meta.Write` already does
- export `storage.SameBackend`, which is how the target decides whether Milvus needs credentials of its own

## Notes for review

`writeMeta` in the backup task and `meta.Write` in the meta package now both encode the same layout rule, because the backup task writes its meta through `metaBuilder` rather than through `meta.Write`. They also disagree on what `backup_meta.json` holds — `meta.Write` puts RBAC, databases and flush messages in it, `buildBackupMeta` five scalars. That is pre-existing, and unifying the two writers seems worth its own PR rather than a detour in this one.

A follow-up, once milvus-io/milvus#50978 is released: point `StrategyAuto` at the snapshot strategy on a server that has the export, and refuse `bulk_flush` / `serial_flush` / `skip_flush` there. `meta_only` stays valid — it collects DDL and touches no data.

End to end this needs milvus-io/milvus#50978, still open.

### DIFF
--- a/core/backup/coll_snapshot_task.go
+++ b/core/backup/coll_snapshot_task.go
@@ -1,0 +1,201 @@
+package backup
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/milvus-io/milvus-proto/go-api/v3/milvuspb"
+	"go.uber.org/zap"
+
+	"github.com/zilliztech/milvus-backup/core/proto/backuppb"
+	"github.com/zilliztech/milvus-backup/internal/client/milvus"
+	"github.com/zilliztech/milvus-backup/internal/log"
+	"github.com/zilliztech/milvus-backup/internal/namespace"
+	"github.com/zilliztech/milvus-backup/internal/taskmgr"
+)
+
+const (
+	_snapshotPollInterval = 3 * time.Second
+
+	// A drop right after the export job goes terminal can still be rejected: DataCoord
+	// releases the pin it took for the copy on its own reconcile tick, not inline with
+	// the state change the poll observed.
+	_snapshotDropRetry    = 10
+	_snapshotDropInterval = time.Second
+)
+
+type collSnapshotTask struct {
+	taskID string
+
+	ns           namespace.NS
+	snapshotName string
+	target       snapshotTarget
+
+	pollInterval time.Duration
+
+	grpc milvus.Grpc
+
+	taskMgr     *taskmgr.Mgr
+	metaBuilder *metaBuilder
+
+	logger *zap.Logger
+}
+
+func newCollSnapshotTask(ns namespace.NS, snapshotName string, target snapshotTarget, args collTaskArgs) *collSnapshotTask {
+	logger := log.L().With(
+		zap.String("task_id", args.TaskID),
+		zap.String("ns", ns.String()),
+		zap.String("snapshot_name", snapshotName))
+
+	return &collSnapshotTask{
+		taskID:       args.TaskID,
+		ns:           ns,
+		snapshotName: snapshotName,
+		target:       target,
+		pollInterval: _snapshotPollInterval,
+		grpc:         args.Grpc,
+		taskMgr:      args.TaskMgr,
+		metaBuilder:  args.MetaBuilder,
+		logger:       logger,
+	}
+}
+
+// Execute freezes the collection into a snapshot, has Milvus copy it into the backup
+// directory, and records where the bundle landed. No bytes move through this process.
+func (st *collSnapshotTask) Execute(ctx context.Context) error {
+	st.logger.Info("start to backup collection with snapshot")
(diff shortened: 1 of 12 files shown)


lentitude2tk (approved):
/LGTM

PR #1134, opened 2026-07-30
restore: add a snapshot path that restores exported bundles

huanghaoyuanhhy:
## Summary

Restores a snapshot format backup by handing the exported bundle back to Milvus, instead of importing binlogs. `restore.NewTask` now routes on `BackupInfo.format`: the binlog format keeps the existing collection task, the snapshot format gets one that submits `RestoreExternalSnapshot` and follows the job to completion.

Milvus owns the whole restore on that path — it creates the collection from the schema in the bundle, restores its indexes and partitions, copies the data in, and refuses to restore over a collection that already exists. That decides the shape of the rest:

- there is no DDL task, so `drop_exist_collection` is honored here, before the job is submitted;
- options that would alter what gets created are refused up front rather than silently dropped: `skip_create_collection`, `meta_only`, `max_shard_num`, `truncate_binlog_by_ts`, `ezk_mapping`, `skip_params`, and per-collection overrides;
- the index options (`restore_index`, `use_auto_index`, `drop_exist_index`) and `use_v2_restore` are only logged, because Milvus restores the bundle's own indexes either way and there is no import job to pick an API for.

Renaming, filtering, RBAC and database restore are unchanged — they run before Milvus is asked for anything.

The metadata URI is rebuilt from the path the backup meta records relative to the backup directory, so a backup moved to another bucket or prefix stays restorable. Building that URI and the `extfs` spec is the same translation the export side already does in the opposite direction, so both now call into `internal/storage`.

## Changes

- `internal/storage/snapshot.go`: `SnapshotURI` and `SnapshotExternalSpec`, moved out of `core/backup` so the restore side shares them
- `core/restore/snapshot_source.go`: where the bundles are read from, and the URI of one collection's metadata within it
- `core/restore/coll_snapshot_task.go`: submits the restore job, polls it, reports progress and drops an existing target when asked
- `core/restore/task.go`: format routing, the option checks above, a `collectionTask` interface so both formats share the filtering and database preparation, and a single server-version check instead of one failure per collection

Related to #1119


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

PR #1135, opened 2026-07-31
backup: split the format axis from the strategy

huanghaoyuanhhy:
## Summary

The strategy enum carried both the flush policy and the artifact choice: `snapshot` was a strategy value (#1133), and the format recorded in the meta was derived from it. Two orthogonal questions were collapsed into one knob, and no strategy but the snapshot-specific one could produce a snapshot backup. Restore already dispatched on the format — the production side should write it as its own axis.

## Changes

- Remove `StrategySnapshot` from the strategy enum and parse map; `strategy=snapshot` now errors with `use --format=snapshot instead`
- Add the `format` axis: `backup.Format` enum, `--format` CLI flag, `CreateBackupRequest.format` proto field, server `toFormat()`
- Each explicit format is a strict pin against the version fork: `snapshot` needs a 3.0+ server, `binlog` is refused on one; `auto` follows the fork but stays on the binlog path until the export lands on master
- The plans compose the two axes at construction: `selectPlan` picks the flush orchestration the strategy asks for and passes it the data-task factory the format resolves to, so every flush plan works under either format; the old `snapshotStrategy` (bulk-flush with a snapshot data task) disappears
- Rename the strategy skeletons to plans: `XxxStrategy` → `XxxPlan`, `selectStrategy` → `selectPlan`, `coll_strategy.go` → `coll_plan.go`

/kind improvement


lentitude2tk (approved):
/LGTM

wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

PR #1137, opened 2026-08-04
backup: default auto format to snapshot on 3.0 servers

huanghaoyuanhhy:
## Summary

Auto used to resolve to the binlog format regardless of the server, leaving --format=snapshot as the only way to get a snapshot backup on a 3.0 server. The export the format depends on has landed, so auto now follows the version fork: a 3.0 or newer server gets the snapshot format, anything older falls back to binlogs.

## Changes

- resolveFormat(): the auto case resolves to the snapshot format when the server reports the Snapshot feature (>= 3.0) and to binlogs otherwise
- --format=binlog stays accepted on 3.0 servers for now: milvus master reports the snapshot feature but has not implemented ExportSnapshot yet, so the explicit pin is the only way to keep producing binlog backups there; revisit when the export lands
- The master CI scenario (same-version-master) pins --format=binlog until the export lands
- Add table-driven TestTask_resolveFormat covering auto on both server lines and the format pins

/kind improvement

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -742,7 +742,10 @@ jobs:
         run: |
           ./milvus-backup check
           ./milvus-backup list
-          ./milvus-backup create -n my_backup --backup_index_extra
+          # Pin binlog until ExportSnapshot lands on the milvus master line: master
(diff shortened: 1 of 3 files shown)


lentitude2tk (approved):
/LGTM

Issue #964, opened 2026-03-02
[Bug]: milvus-backup does not support backing up from local storage to MinIO for Milvus standalone？


QunLeLZ:
### Milvus-Backup Version

v0.5.11

### Can you reproduce this issue on the latest version?

Not tested on the latest version

### Current Behavior

I am using Milvus standalone 2.5.19 with local storage.
When I try to use milvus-backup to back up data from local storage and restore it to a new Milvus cluster with MinIO, milvus-backup does not support backing up local data to MinIO.

### Steps To Reproduce

```markdown

```

### Environment

```markdown
Milvus version: 2.5.19 (standalone)
milvus-backup version: v0.5.11
```

huanghaoyuanhhy:
Hi @QunLeLZ, we'd like to support this scenario. Could you share the exact error message/logs and your `backup.yaml` config (with credentials redacted)? This will help us implement and verify this capability.

QunLeLZ:
Thanks for your reply!
I configured backup.yaml like this:
```
# Configures the system log output.
log:
  level: info # Only supports debug, info, warn, error, panic, or fatal. Default 'info'.
  console: true # whether print log to console
  file:
    filename: "logs/backup.log"

# milvus proxy address, compatible to milvus.yaml
milvus:
  address: xxxxxxxx
  port: 31953
  user: ""
  password: ""

  # tls mode values [0, 1, 2]
  # 0 is close, 1 is one-way authentication, 2 is mutual authentication
  tlsMode: 0
  # tls cert path for validate server, will be used when tlsMode is 1 or 2
  caCertPath: ""
  serverName: ""
  # mutual tls cert path, for server to validate client.
  # Will be used when tlsMode is 2
  # for backward compatibility, if not set, will use tlsmode 1.
  # WARN: in future version, if user set tlsmode 2, but not set mtlsCertPath, will cause error.
  mtlsCertPath: ""
  mtlsKeyPath: ""

  # Milvus replicate msg channel name, default is by-dev-replicate-msg
  rpcChannelName: "by-dev-replicate-msg"

# Related configuration of minio, which is responsible for data persistence for Milvus.
minio:
  storageType: "local"
  rootPath: "/opt/data/hdd/misc/milvus/data/"

  # Backup storage configs, the storage you want to put the backup data
  backupStorageType: "minio" # support storage type: local, minio, s3, aws, gcp, ali(aliyun), azure, tc(tencent)
[...]

PR #1138, opened 2026-08-05
ci: test backup and restore of a local-storage milvus

huanghaoyuanhhy:
## Summary

Milvus standalone with `COMMON_STORAGETYPE=local` keeps its data on the local filesystem at `localStorage.path` (default `/var/lib/milvus/data`). The storage plumbing already reads this — `milvus.storage.provider: local` with a `rootPath` pointing at the directory on disk resolves through the `LocalClient`, because every storage key is rebuilt from the configured root path rather than read back from etcd metadata. What was missing is the configuration and proof, so backing up a local-storage Milvus to MinIO was effectively undocumented and untested.

Part of #964

## Changes

- add `configs/backup-local-milvus.yaml`, a sample backing a local-storage Milvus up to MinIO
- add `deployment/standalone/docker-compose.local.yaml`, a compose override that runs the standalone with `COMMON_STORAGETYPE=local`
- add a CI job that backs a local-storage Milvus up to MinIO and restores it into a fresh MinIO-backed Milvus, verifying the data round-trips
- note the local provider in `configs/backup.yaml` and list the new sample in the README

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -263,6 +263,176 @@ jobs:
         run: |
           python example/verify_data.py
 
+  test-backup-restore-local-storage:
+    needs: unit-test-go
+    # Milvus standalone with COMMON_STORAGETYPE=local keeps its data on the
+    # local filesystem (localStorage.path, default /var/lib/milvus/data). This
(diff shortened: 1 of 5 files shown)


lentitude2tk (approved):
/LGTM

PR #1139, opened 2026-08-05
storage: support gcp and gcpnative providers in the snapshot path

huanghaoyuanhhy:
## Summary

Part of #1119: the snapshot export/restore path only mapped the s3-family providers, so any backup whose bucket is GCP storage was refused with `milvus snapshots do not support gcp storage` before any snapshot work started.

This wires the two GCP providers Milvus accepts in `extfs.cloud_provider`:

- `gcp` (S3-family, minio client against `storage.googleapis.com`) -> maps to `"gcp"`, authorized with `use_iam` (workload identity) or access key pair, matching how the backup pod is configured today.
- `gcpnative` (native GCS) -> maps to `"gcpnative"`, emits a `gcs://` uri (the scheme decides the provider family on the Milvus side), and carries the service-account json as `credential_json`.

## Changes

- `snapshotCloudProvider`: add `gcp` and `gcpnative` cases.
- `SnapshotURI`: emit `gcs://<bucket>/<key>` for native GCS.
- `SnapshotExternalSpec`: read the `GCPCredJSON` credential file and send it as `credential_json`.
- Tests for the new URI shapes and extfs specs.

## Notes

- Related to #1119
- Depends on the Milvus-side external snapshot export/restore flow (milvus-io/milvus#50978) being present on the server.
- Azure is wired in a companion PR.

### DIFF
--- a/internal/storage/snapshot.go
+++ b/internal/storage/snapshot.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"encoding/json"
 	"fmt"
+	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/LGTM

PR #1141, opened 2026-08-05
restore: allow non-binding max_shard_num on snapshot restores

huanghaoyuanhhy:
## Summary
Restoring a snapshot-format backup currently rejects the `max_shard_num` option outright, even when the requested cap is no smaller than the bundle's shard count and therefore changes nothing. This PR relaxes the check to be per-collection: `max_shard_num` is refused only when it would actually cap shards, which the snapshot path cannot do because Milvus creates the collection from the schema in the bundle.

## Changes
- `checkSnapshotSupport` no longer refuses `max_shard_num` for snapshot-format restores
- `collSnapshotTask` gains a per-collection check that fails the collection only when `max_shard_num` would bind, and logs an info line when the cap is accepted as a no-op
- tests cover both the satisfied-cap and the binding-cap paths

/kind improvement

### DIFF
--- a/core/restore/coll_snapshot_task.go
+++ b/core/restore/coll_snapshot_task.go
@@ -24,8 +24,9 @@ type collSnapshotTask struct {
 	collBackup *backuppb.CollectionBackupInfo
 	targetNS   namespace.NS
 
-	source    snapshotSource
-	dropExist bool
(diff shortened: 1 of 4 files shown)


lentitude2tk (approved):
/LGTM

PR #1142, opened 2026-08-05
restore: support restoring into a local-storage milvus

huanghaoyuanhhy:
## Summary

Related to #964

Milvus standalone with `COMMON_STORAGETYPE=local` keeps its data on the local filesystem (`localStorage.path`). Backing one up and restoring into a remote-backed Milvus works (see #1138); restoring into a **local-storage target** did not: restore wrote its temp files under a bucket-relative key (the `LocalClient` treats keys as absolute paths, so they landed in the tool's working directory) and handed Milvus bucket-relative import paths that its `LocalChunkManager` cannot resolve (it globs the path directly and never prepends its own root).

This adds the optional `milvus.storage.localPath` config — the path the Milvus process resolves its `localStorage.path` at when it differs from `rootPath`, e.g. a container bind-mount — and makes restore local-aware: temp files are written under `rootPath`, and the import paths passed to Milvus are absolute under `localPath` (trailing slash kept, since the local chunk manager lists a prefix by globbing it).

## Changes
- cfg: add optional `milvus.storage.localPath`, rejected for non-local providers
- restore: detect a local target and resolve restore temp keys (`rootPath`) and bulk-insert paths (`localPath`) against the local directory
- CI: the local-storage job now restores into a second local-storage Milvus instead of a MinIO-backed one
- configs: document `localPath` in the local storage samples

### DIFF
--- a/.github/workflows/main.yaml
+++ b/.github/workflows/main.yaml
@@ -268,13 +268,16 @@ jobs:
     # Milvus standalone with COMMON_STORAGETYPE=local keeps its data on the
     # local filesystem (localStorage.path, default /var/lib/milvus/data). This
     # job proves milvus-backup can read that directory directly and back it up
-    # to MinIO, then restores into a fresh MinIO-backed Milvus (issue #964).
+    # to MinIO, then restores it into a fresh Milvus backed by MinIO or by local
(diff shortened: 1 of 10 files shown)


lentitude2tk (approved):
/LGTM

PR #1143, opened 2026-08-06
server: log http error responses

huanghaoyuanhhy:
## Summary
Log an error at Error level when an HTTP handler returns a non-success response, so failed backup/delete/get/restore requests leave a trace in the server logs instead of only appearing in the response body.

## Changes
- create: log create backup failure in run()
- delete: log delete backup failure in run()
- get_backup: log get backup failure in get()
- restore: log restore backup failure in run()
- restore_secondary: log restore secondary failure in run()

/kind improvement


lentitude2tk (approved):
/LGTM

PR #1144, opened 2026-08-06
proto: regenerate backup.pb.go via gen_proto.sh

huanghaoyuanhhy:
/kind improvement

## Summary
Regenerate `core/proto/backuppb/backup.pb.go` with `scripts/gen_proto.sh`. The script post-processes the generated code to drop `omitempty` from several json tags, but the committed file had drifted: the last regeneration (#1132/#1135) ran `protoc` directly and skipped the script's `sed` steps.

## Changes
- Regenerate backup.pb.go via gen_proto.sh, restoring the script's documented post-processing
- `size`, `has_index`, `data`, `progress`, `state_code` (plus `log_size`, `meta_size` via the script's substring substitution) now always serialize in JSON; zero values show as `0`/`false` instead of being omitted
- An empty binlog collection now reports `"size": 0` in backup responses instead of omitting the field
- Wire format is unchanged (only json tags differ); existing backups still parse


lentitude2tk (approved):
/LGTM

PR #982, opened 2026-03-11
restore: support per-collection shardNum and description override

huanghaoyuanhhy:
## Summary
- Add `RestoreCollectionOverride` message to proto, nested inside `RestoreCollectionMapping`, allowing callers to specify `shardNum` and `description` per collection during restore via `restorePlan`
- Override priority: `collOverride > maxShardNum cap > backup value`

## Changes
- Proto: add `RestoreCollectionOverride` message with `shardNum` and `description` fields, add `override` field to `RestoreCollectionMapping`
- `core/restore/task.go`: add `CollOverride` struct and `Plan.CollOverrides` map; thread override to per-collection tasks
- `core/restore/coll_task.go`: propagate `CollOverride` from args to `collDDLTask`
- `core/restore/coll_ddl_task.go`: add `description()` method; update `shardNum()` to prioritize override
- `core/server/restore.go`: add `newCollOverridesFromPlan()` to extract overrides from `RestorePlan`
- Unit tests: 13 new test cases covering override priority, fallback, and extraction logic
- E2E test: `test_milvus_restore_with_collection_override` verifying shard_num/description override with data integrity

### API usage example
```json
{
  "backup_name": "my_backup",
  "restorePlan": {
    "mapping": [{
      "source": "default",
      "target": "default",
      "colls": [{
        "source": "my_coll",
        "target": "my_coll_restored",
        "override": {
          "shardNum": 2,
          "description": "new description"
        }
      }]
    }]
  }
}
```

/kind improvement

PR #1147, opened 2026-08-10
restore: support collection description override on snapshot restores

huanghaoyuanhhy:
## Summary
The snapshot restore path hands the whole restore to Milvus, which creates the collection from the bundle's schema, so the per-collection description override that the binlog path has honored since #982 was refused up front for snapshot backups. This PR applies it instead: once the restore job completes, the target collection's description is rewritten with an AlterCollection carrying the `collection.description` property.

## Changes
- `core/restore/task.go`: `checkSnapshotSupport` now only refuses `shardNum` overrides; a `description` override is threaded into the snapshot collection task
- `core/restore/coll_snapshot_task.go`: after the restore job completes, alter the collection with the `collection.description` property when an override is set
- tests: snapshot task override / empty / alter-failure cases, and the snapshot support check

Related to #1119

### DIFF
--- a/core/restore/coll_snapshot_task.go
+++ b/core/restore/coll_snapshot_task.go
@@ -6,7 +6,9 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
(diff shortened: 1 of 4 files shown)


lentitude2tk (approved):
/LGTM

PR #1149, opened 2026-08-11
storage: unify object iterator into a single Next method

huanghaoyuanhhy:
## Summary

Refactor the object-listing iterator in `internal/storage` from the two-phase `HasNext() bool` + `Next() (ObjectAttr, error)` contract to a single pull method `Next(ctx) (ObjectAttr, bool, error)`, following the `(*sql.Rows).Next()` idiom. `ok=true` means a real object was returned, `ok=false` means the listing is exhausted, and `err` is non-nil only on a real failure.

This removes the side-effecting `HasNext` (the Azure iterators fetched pages inside it), makes exhaustion vs failure unambiguous across providers, and threads the caller's `ctx` into Azure page fetches so iteration is cancellable. It also fixes a latent bug where the Azure flat iterator silently swallowed pagination errors, letting copy/delete/verify tasks report success on an incomplete listing.

## Changes

- `storage`: collapse `ObjectIterator` to `Next(ctx) (ObjectAttr, bool, error)`; drop the pre-read double-buffering in the minio and gcp-native iterators
- `storage`: unify the flat and hierarchy Azure iterators into a shared generic `pageIterator[T]`, using the caller context for page fetches
- `storage`: update the five call sites (`ListPrefixFlat`, `DeletePrefix`, `Exist`, `CopyPrefixTask`, `VerifyPrefixTask`) to the new loop idiom
- `test`: update the local and mock iterator tests; add Azure pagination-error and empty-page contract tests

/kind improvement

### DIFF
--- a/internal/storage/azure.go
+++ b/internal/storage/azure.go
@@ -347,127 +347,53 @@ func (a *AzureClient) UploadObject(ctx context.Context, i UploadObjectInput) err
 	return nil
 }
 
-type AzureObjectFlatIterator struct {
-	cli *AzureClient
-
-	pager *runtime.Pager[azblob.ListBlobsFlatResponse]
+// pageIterator is a pull-based iterator over the Azure pager API. It fetches
+// one page at a time, skips pages that yield no objects, and surfaces pager
+// errors through Next. Both flat (recursive) and hierarchy (delimiter) listings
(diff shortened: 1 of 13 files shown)


lentitude2tk (approved):
/LGTM

PR #1150, opened 2026-08-11
storage: stop background listing goroutines on iterator close

huanghaoyuanhhy:
## Summary

minio-go's `ListObjects` spawns a background goroutine that feeds a channel and only returns once its context is cancelled and the channel is drained. The iterators returned by `ListPrefix` left that goroutine running when a consumer stopped early: `ListPrefixFlat`, `DeletePrefix` and `Exist` in `funcs.go` (and `CopyPrefixTask`/`VerifyPrefixTask`) returned on error or after a single read without stopping it, leaking a goroutine that keeps listing the whole prefix until the creation context is cancelled.

The same functions also abandoned errgroup workers on early return: `DeletePrefix`, `CopyPrefixTask.Execute` and `CopyObjectsTask.Execute` spawned delete/copy goroutines via `errgroup`, then returned before `g.Wait()` on an iterator or semaphore-acquire error, leaving the workers running with a context that never gets cancelled (errgroup only cancels its derived context when a worker errors or `Wait` returns).

The iterator now owns its lifecycle and errgroup workers are joined on every exit path. `ObjectIterator` embeds `io.Closer`, every consumer defers `iter.Close()`, and the minio iterator cancels its listing context and drains the channel so the goroutine deterministically exits. Azure, GCP and local iterators are pull-based, so their `Close` is a no-op. The errgroup consumers capture the loop error, break out, cancel the in-flight workers when bailing early, and join them through a single `g.Wait()`; a `defer cancel()` keeps `go vet`'s lostcancel check satisfied on the happy path.

## Changes

- add `Close` to `ObjectIterator` (`io.Closer`) with a documented caller contract
- minio: cancel the listing context and drain on `Close` so the goroutine exits even on early stop
- `defer iter.Close()` in `ListPrefixFlat`, `DeletePrefix`, `Exist`, `CopyPrefixTask` and `VerifyPrefixTask`
- `DeletePrefix`, `CopyPrefixTask.Execute` and `CopyObjectsTask.Execute`: single `g.Wait()` on every exit path, cancelling in-flight workers first when the loop bails early
- tests: assert each consumer closes the iterator; a transport-level test that `Close` stops a live truncated listing; regression tests that early-return paths cancel and join in-flight deletes/copies (verified to fail fast without the fixes)

/kind improvement

### DIFF
--- a/internal/storage/azure.go
+++ b/internal/storage/azure.go
@@ -388,6 +388,10 @@ func (p *pageIterator[T]) Next(ctx context.Context) (ObjectAttr, bool, error) {
 	}
 }
 
+// Close is a no-op: the Azure pager fetches pages synchronously on Next and
+// keeps no background goroutine.
+func (p *pageIterator[T]) Close() error { return nil }
+
 type AzureObjectFlatIterator struct {
 	pageIterator[azblob.ListBlobsFlatResponse]
 }
--- a/internal/storage/client.go
+++ b/internal/storage/client.go
@@ -115,7 +115,13 @@ func (c CredentialType) String() string {
 // ObjectIterator yields objects under a prefix. Each Next call returns the
 // next object, or ok=false once the listing is exhausted. err is non-nil only
 // on a real failure; exhaustion is not an error.
+//
+// The caller must Close the iterator when done, even after exhaustion or an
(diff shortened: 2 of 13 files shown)


lentitude2tk (approved):
/LGTM

PR #1151, opened 2026-08-12
storage: read azure objects via download stream

huanghaoyuanhhy:
## Summary
Replace the hand-rolled Azure reader with the azure-sdk `DownloadStream`, which fetches the whole object in a single GET and streams its body, instead of issuing one ranged GET per `Read` call via `DownloadBuffer`.

## Changes
- Read objects with a single `DownloadStream` GET instead of per-chunk `DownloadBuffer`
- Drop the extra `GetProperties` call — length comes from the download response
- Use the SDK's `NewRetryReader` so a mid-stream failure resumes via a ranged request
- Remove the now-unused `AzureReader` type and the `io` import

/kind improvement

### DIFF
--- a/internal/storage/azure.go
+++ b/internal/storage/azure.go
@@ -5,7 +5,6 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
-	"io"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


lentitude2tk (approved):
/LGTM

PR #1160, opened 2026-08-14
backup: accept canonicalized snapshot metadata uri

### LINKED ISSUE #1159: Snapshot backup fails on canonicalized metadata URI
## Problem

Snapshot export can succeed while the backup task fails afterward when Milvus returns the exported metadata URI in a canonical transport form.

For example, milvus-backup sends:

```text
minio://s3.us-west-2.amazonaws.com:443/bucket/backup/bundle
```

Milvus may return:

```text
https://s3.us-west-2.amazonaws.com/bucket/backup/bundle/exports/.../metadata.json
```

These identify the same endpoint, bucket, and object prefix, but `snapshotTarget.metadataPath` currently uses a raw string prefix comparison and rejects the result.

## Expected behavior

Validate the storage identity and object-key boundary semantically, accepting equivalent endpoint URI forms while continuing to reject a different endpoint, bucket, or sibling prefix.

huanghaoyuanhhy:
Fixes #1159

## What changed

- parse snapshot target and exported metadata URIs into endpoint, bucket, and object-key components
- treat transport-normalized forms such as `minio://host:443` and `https://host` as the same endpoint
- retain strict bucket, endpoint, path-boundary, traversal, query, fragment, and credential checks
- add regression coverage for canonicalized endpoints and rejected sibling prefixes/endpoints

## Tests

```text
GOCACHE=/private/tmp/milvus-backup-go-cache go test ./core/backup
```

### DIFF
--- a/core/backup/snapshot_target.go
+++ b/core/backup/snapshot_target.go
@@ -2,6 +2,8 @@ package backup
 
 import (
 	"fmt"
+	"net"
+	"net/url"
 	"path"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/LGTM

PR #1164, opened 2026-08-17
backup: omit snapshot uri endpoint on the same backend as milvus

### LINKED ISSUE #1163: Snapshot backup fails when milvus-backup and Milvus reach the same storage through different endpoints
## Problem

Snapshot-format backup fails on Milvus master whenever milvus-backup and the Milvus server reach the same object storage through different endpoint names — which is the normal case when one side runs inside a container network and the other outside it.

milvus-backup builds the ExportSnapshot target URI from its own storage config, e.g.:

```text
minio://localhost:9000/a-bucket/backup/my_backup/bundle
```

while Milvus itself is configured with `minio:9000`, the in-network name of the same MinIO. Milvus compares the URI endpoint against its own storage endpoint as strings, treats the mismatch as a foreign cross-bucket target, and rejects the export:

```text
backup: export snapshot: client: export snapshot failed: ... server-side cross-bucket copy is unsupported for this provider/endpoint pair; cross-service copy would require streaming (unsupported): invalid parameter
```

Observed in CI: every master-latest job of https://github.com/zilliztech/milvus-backup/pull/1162 fails this way, e.g. https://github.com/zilliztech/milvus-backup/actions/runs/32011427612/job/95332476332

Adding the alias to Milvus `dataCoord.snapshot.crossBucketEndpointAllowlist` passes the check but is not a fix: the export client is then built against the URI endpoint (`localhost:9000`), which the Milvus container cannot connect to either.

## Expected behavior

A backup kept in the same storage backend as Milvus should not depend on both sides spelling the endpoint identically. When the backup storage is the same backend, milvus-backup should omit the endpoint from the snapshot URI (`s3://<bucket>/<key>`) and let Milvus resolve the bucket through its own storage config; an optional `milvusAddress`/`milvusPort` override should cover deployments that need to pin the Milvus-view endpoint explicitly.

## Environment

- milvus-backup: master (0902e30)
- Milvus: master-latest (snapshot export is master-only; 2.6 jobs use the binlog format and are unaffected)

huanghaoyuanhhy:
## Summary

Fixes #1163

Snapshot-format backup on Milvus master fails whenever milvus-backup and the Milvus server reach the same storage through different endpoint names (e.g. `localhost:9000` from the host vs `minio:9000` inside a container network): milvus-backup wrote its own view of the endpoint into the ExportSnapshot target URI, and Milvus rejects the string mismatch as a foreign cross-bucket target.

A snapshot export/restore is executed by Milvus, so the URI endpoint must be one Milvus can reach. When the backup storage is the same backend as Milvus, that endpoint is redundant at best and an unresolvable alias at worst — so it is now omitted from the URI (`s3://<bucket>/<key>`), and Milvus resolves the bucket through its own storage config. No endpoint comparison happens on that path at all.

## Changes

- `storage.SnapshotStoreURI`: omit the endpoint for same-backend S3-family storage; fall back to the previous behavior otherwise (Azure has no endpoint-less form, native GCS never carries one)
- new `backup.storage.milvusAddress`/`milvusPort` config to pin the Milvus-view endpoint explicitly, following the `localPath` precedent; wins over omission when set
- `snapshotTarget.sameStorage`: accept the endpoint-style metadata URI Milvus rebuilds from its resolved storage config when the target named only a bucket, as long as bucket and provider family match
- config validation, sample config docs, and tests for the above

### DIFF
--- a/configs/backup.yaml
+++ b/configs/backup.yaml
@@ -118,6 +118,15 @@ backup:
     bucketName: "a-bucket"
     rootPath: "backup"
 
+    # The endpoint the Milvus server itself uses to reach this storage, when
+    # it differs from address/port above: a container port mapping, a private
+    # link, or an internal DNS name gives one store two endpoints. Snapshot
+    # URIs handed to Milvus must name the endpoint Milvus resolves. Left
+    # empty, a backup kept in the same backend as Milvus omits the endpoint
+    # from those URIs entirely and lets Milvus use its own storage config.
+    # milvusAddress: ""
+    # milvusPort: 0
+
     auth:
       type: static
       accessKeyID: minioadmin
--- a/core/backup/snapshot_target.go
+++ b/core/backup/snapshot_target.go
@@ -28,7 +28,7 @@ type snapshotTarget struct {
 }
 
 func newSnapshotTarget(milvusCfg, backupCfg storage.Config, backupDir string) (snapshotTarget, error) {
-	uri, err := storage.SnapshotURI(backupCfg, mpath.BackupBundleDir(backupDir))
+	uri, err := storage.SnapshotStoreURI(milvusCfg, backupCfg, mpath.BackupBundleDir(backupDir))
 	if err != nil {
 		return snapshotTarget{}, err
 	}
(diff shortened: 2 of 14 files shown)


lentitude2tk (approved):
/LGTM

wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

Issue #1167, opened 2026-08-18
[Bug]: restore secondary broadcasts CreateIndex with FieldID 0 when the backup was taken without --backup_index_extra


czs007:
### Milvus-Backup Version

main (a9d6043) and 0.5.16

### Can you reproduce this issue on the latest version?

Yes, it also happens on the latest version

### Output of `milvus-backup check`

```markdown
(not relevant — both storages fine)
```

### Current Behavior

When the backup was taken **without** `--backup_index_extra`, `restore secondary` broadcasts every CreateIndex with `FieldID: 0`. `IndexInfo.field_id`, `type_params`, `index_params` and `create_time` in the backup meta are only populated by the index-extra task (read from etcd); without it the meta carries `field_name`, `index_name`, `index_type`, `params` and `index_id` only. `core/restore/secondary/coll_ddl_task.go:createIndex` copies `index.GetFieldId()` straight into the message, so the target ends up with an index on field 0:

- the import job finishes importing (`import job import done`, sort compaction created) and then stays in `IndexBuilding` forever; the client shows `[state=Importing] [progress=80]` until timeout;
- DataCoord index tasks fail with `parse magic number failed, expected: 16775868, actual: 827474256` (0x31524150 = "PAR1": the index node opens the storage-v2 packed column file of field 0 as a v1 binlog);
- `DescribeIndex` on the restored collection fails with `failed to get collection field: 0`.

The same backup restores fine with plain `restore`, which sends `FieldName` through the CreateIndex RPC and lets the server resolve it.

### Expected Behavior

`restore secondary` should not depend on `--backup_index_extra`. When `field_id` is 0, resolve the field from the collection schema by `field_name`; when `index_params` / `type_params` are empty, rebuild them from `params` and the schema field's `type_params`.

### Steps To Reproduce

```markdown
1. Collection with a vector index on A. `milvus-backup create -n xb4 --filter default.<coll>` (no --backup_index_extra).
2. Confirm meta: `index_infos[0]` has no `field_id`.
3. Apply A→B config on B; `restore secondary -n xb4 ...`
4. Client stays at `[state=Importing] [progress=80]`; B DataCoord: `parse magic number failed ...`; DescribeIndex: `failed to get collection field: 0`.
5. Repeat with `--backup_index_extra` → completes.
```

### Environment

```markdown
- Milvus: v2.6.18, milvus-backup 0.5.16 / main
```

### Anything else?

[...]

Issue #1166, opened 2026-08-18
[Bug]: restore secondary never completes when the backup bucket differs from the Milvus bucket (no binlog to import, client loops on empty state)


czs007:
### Milvus-Backup Version

main (a9d6043) and 0.5.16

### Can you reproduce this issue on the latest version?

Yes, it also happens on the latest version

### Output of `milvus-backup check`

```markdown
(both storages reachable; milvus.storage bucket = milvus-bucket, backup.storage bucket = backup-only, same MinIO endpoint)
```

### Current Behavior

`restore secondary` never completes when the backup lives in a different bucket than the target cluster's own storage. The tool logs

```
["bulk insert task state"] [job_id=...] [state=] [progress=0]
```

every 3 s forever, and the target's DataCoord logs, for every Import message,

```
[broadcaster/ack_callback_scheduler.go:308] ["failed to call message ack callback, wait for retry..."]
  [message="{type=Import,...}"] [error="no binlog to import, input=[id:1 paths:\"backup/<name>/binlogs/insert_log/<coll>/<part>/<seg>/\"]: importing data failed"]
```

Root cause: `core/restore/secondary/coll_dml_task.go` builds the Import message paths directly from the backup directory (`buildBackupPartitionDir` → `mpath.BackupInsertLogDir(dmlt.backupDir, …)`) and broadcasts them as-is. DataCoord resolves `backup=true` import paths only against **its own** storage (`ListBinlogImportRequestFiles` with `s.meta.chunkManager`), so a key that only exists in the backup bucket lists as empty. The plain restore path already handles this (`collTask.copyAndRewriteDir` copies binlogs into `restore-temp-<task>-<db>-<coll>/` in the Milvus bucket when the buckets differ); the secondary path has no equivalent — the `secondary` package never receives a `MilvusStorage` client at all.

Two consequences worth noting:

- The retry never ends: the broadcaster keeps re-issuing the same Import messages (2,142 retries in ten minutes in one field log), so the broadcast queue on the target accumulates dead tasks.
- The client cannot see the failure: `RestfulClient.GetBulkInsertState` only checks the HTTP status. `POST /v2/vectordb/jobs/import/describe` returns HTTP 200 with `{"code":2100,"message":"import job does not exist ..."}` because the job was never created, so `resp.Data.State` is `""`, which matches neither `Failed` nor any done state, and the poll loop in `coll_dml_task.go` runs until the context is cancelled.

### Expected Behavior

[...]

PR #1168, opened 2026-08-18
fix(restore secondary): stage binlogs into milvus storage across buckets

czs007:
issue: #1166

## Problem

`restore secondary` never imports when the backup lives in a different bucket/backend from Milvus.

The secondary DML task builds Import message paths straight from the backup directory and broadcasts them. DataCoord resolves `backup=true` import paths only against its own storage (`ListBinlogImportRequestFiles` with `s.meta.chunkManager`), so when `backup.storage` differs from `milvus.storage` every job fails with `no binlog to import`, the broadcaster retries it forever, and the client loops on `[state=] [progress=0]` (describe returns HTTP 200 with `code:2100` for a job that was never created). The plain restore path handles this in `collTask.copyAndRewriteDir`; the `secondary` package never even receives a Milvus storage client.

## Change

- `secondary.TaskArgs` gains `MilvusStorage`; the CLI (`cmd/restore/secondary.go`) and REST (`/restore_secondary`) entry points construct it.
- `collDMLTask.stageBatch` copies each partition dir into `restore-temp-<task>-<db>-<coll>/` in the Milvus storage (`CopyPrefixTask` + `VerifyPrefixTask`, same rule as `copyAndRewriteDir`: skipped when same bucket, same provider and not streaming), rewrites the batch, and the task removes the staging prefix on completion.

Same-bucket behaviour is unchanged.

## Verification

Milvus v2.6.18 (helm 5.0.22), two clusters on one MinIO; `minio.bucketName: milvus-bucket`, `minio.backupBucketName: backup-only`.

| | before | after |
|---|---|---|
| restore secondary, cross-bucket | `no binlog to import` ×26 on DataCoord, client `[state=] [progress=0]` until timeout | `Importing 10→40→70→80 → Completed 100`, ~15 s, staging prefix deleted |
| rows / collection ID | — | 2500/2500, IDs equal, target Loaded |
| CDC increment after applying source config | — | +500 rows arrive |

Same-bucket restore (existing path) re-run: unchanged, ~25 s.

## Note on the split

This PR previously also carried the `createIndex` field-inference change for #1167. Per @huanghaoyuanhhy's review it has been split out — this PR is now the cross-bucket staging fix only. #1167 will be handled separately.

### DIFF
--- a/cmd/restore/secondary.go
+++ b/cmd/restore/secondary.go
@@ -62,6 +62,11 @@ func (o *secondaryOption) toArgs(params *v2.Config) (secondary.TaskArgs, error) 
 		return secondary.TaskArgs{}, fmt.Errorf("read backup meta: %w", err)
 	}
 
+	milvusStorage, err := storage.NewMilvusStorage(context.Background(), params)
+	if err != nil {
+		return secondary.TaskArgs{}, fmt.Errorf("create milvus storage: %w", err)
+	}
+
 	return secondary.TaskArgs{
 		TaskID: uuid.NewString(),
 
@@ -72,6 +77,7 @@ func (o *secondaryOption) toArgs(params *v2.Config) (secondary.TaskArgs, error) 
 		Params:        params,
 		BackupDir:     backupDir,
 		BackupStorage: backupStorage,
+		MilvusStorage: milvusStorage,
 
 		TaskMgr: taskmgr.DefaultMgr(),
 	}, nil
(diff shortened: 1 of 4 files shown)


huanghaoyuanhhy:
/lgtm cancel

Withdrawing the lgtm for now — please see my comment on #1167. The cross-bucket staging fix (#1166) looks right, but the createIndex field-inference part conflicts with the design contract that secondary backups must be taken with `--backup_index_extra`. Could you split this PR so it carries only the staging fix? The index-side handling (fail-fast validation) can come as a follow-up.

PR #1173, opened 2026-08-20
restore: apply skip_params on snapshot restores via delete_keys

### LINKED ISSUE #1174: restore: skip_params cannot be used with a snapshot format backup
Restoring a snapshot-format backup with non-empty `skip_params` fails:

```
restore: skip_params cannot be used with a snapshot format backup
```

Since the backup side auto-selects the snapshot format on Milvus 3.0+ servers, this blocks any restore that passes `skip_params` — even when the skipped keys never appear in the backup and there is nothing to skip.

`skip_params` should be honored on the snapshot path instead of refused.

huanghaoyuanhhy:
close #1174

## Problem

The snapshot restore path refuses any non-empty `skip_params`:

```
restore: skip_params cannot be used with a snapshot format backup
```

Cross-type restores (e.g. serverless to dedicated) rely on `skip_params` to drop source-cluster overrides like `mmap.enabled`, so they cannot restore snapshot-format backups at all. Since the backup side auto-selects the snapshot format on Milvus 3.0+, every cross-type restore of a 3.0 backup hits this. It fires even when the skipped keys are absent from the backup entirely, which is the common case.

## Change

Instead of refusing, apply `skip_params` after the snapshot restore job completes, using the `delete_keys` field of the Alter RPCs:

- collection properties -> `AlterCollection`
- field type params -> `AlterCollectionField` (only fields carrying the keys)
- index params + field index params -> `AlterIndex` (only indexes carrying the keys)

Deleting an override makes each key fall back to the target cluster's own default, which is exactly the semantics "don't write it at creation time" has on the binlog path — and it never needs to know what that default is.

What gets deleted is filtered against the schema saved in the backup meta (`schema.properties`, `fields[].type_params`, `index_infos[].params`), so a skip list naming params the collection never set is a quiet no-op rather than an error.

## Notes

- `delete_keys` on these Alter requests exists since Milvus 2.5, and the snapshot format already requires 3.0+ on both backup and restore sides, so no version gating is needed.
- grpc client: added `DropCollectionProperties` / `DropCollectionFieldProperties` / `DropIndexProperties`; mocks regenerated with mockery.
- Tests: flipped the `checkSnapshotSupport` case for skip_params from refused to accepted, and added snapshot-task tests covering selective drop, absent keys, and drop failure propagation.

### DIFF
--- a/core/restore/coll_snapshot_task.go
+++ b/core/restore/coll_snapshot_task.go
@@ -3,12 +3,14 @@ package restore
 import (
 	"context"
 	"fmt"
+	"sort"
 	"strconv"
 	"time"
 
 	"github.com/milvus-io/milvus-proto/go-api/v3/commonpb"
 	"github.com/milvus-io/milvus-proto/go-api/v3/milvuspb"
 	"github.com/milvus-io/milvus/pkg/v3/common"
+	"github.com/samber/lo"
 	"go.uber.org/zap"
 
 	"github.com/zilliztech/milvus-backup/core/proto/backuppb"
@@ -30,6 +32,7 @@ type collSnapshotTask struct {
 	dropExist    bool
 	maxShardNum  int32
 	descOverride string
+	skipParams   SkipParams
 
 	pollInterval time.Duration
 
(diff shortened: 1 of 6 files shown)


lentitude2tk (approved):
/LGTM

Issue #1170, opened 2026-08-20
[Enhancement]: add purpose presets to create (e.g. --for secondary) so backup options match the restore contract


huanghaoyuanhhy:
### What would you like to be added?

A purpose preset on `create`, e.g. `milvus-backup create -n b1 --for secondary`, that expands to the parameter combination the corresponding restore path requires, instead of asking users to pick individual flags and get the combination right themselves.

`--for secondary` should pin the contract that `restore secondary` depends on:

- `backup_index_extra=true` — the only reliable source of `field_id` / `index_params` / `type_params` / `create_time` in the backup meta (see #1167).
- The etcd-backed dynamic-field backup forced on rather than silently skipped when etcd is unconfigured (today `backupCollDynField` skips silently — the same class of silent gap, see #1013).
- `format` constrained to what the secondary replay path supports.
- Fail fast at backup time when etcd is not configured or unreachable, instead of producing a backup that looks successful but can never serve a secondary restore.

Semantics of the preset mechanism:

- A preset expands into concrete flag/option values; the expansion is printed at create time and recorded in the backup meta for auditability.
- An explicit flag conflicting with the preset is an error (e.g. `--for secondary --backup_index_extra=false`), never a silent override.
- Both entry points get it: the CLI flag and an equivalent field on the REST create request.
- The mechanism should generalize to other purposes later (`--for clone`, `--for archive`, ...).

### Why is this needed?

#1167 shows the failure mode: a default `create` produces a backup that cannot serve `restore secondary`, and the user only finds out after the import sits at 80% progress for hours. The underlying contract ("backups for secondary restore must carry the etcd-sourced index/dynamic-field info") is currently implicit and lives only in documentation, so it is easy to violate and drifts as more etcd-backed extras are added. A preset makes the contract executable: one flag, validated up front.

This complements, not replaces, the fail-fast validation proposed in #1167: validation on `restore secondary` protects existing backups that were already taken without the extra info, while the preset ensures future backups are created correctly in the first place.

### Anything else?

Open questions for the implementation:

- What else belongs in the `secondary` preset — the exact `format` constraint, and whether `rbac` should default on for DR scenarios.
- Flag naming: decided — `--for secondary` (reads as "a backup for secondary restore").

Related: #1167, #1013.

PR #1175, opened 2026-08-20
restore: refuse a secondary restore whose backup has no index extra info

czs007:
issue: #1167

Follow-up to #1168.

## Problem

`IndexInfo` splits into two groups by source:

| source | fields |
|---|---|
| DescribeIndex (always present) | `field_name` `index_name` `index_type` `params` `index_id` |
| etcd, only with `--backup_index_extra` | `field_id` `create_time` `type_params` `index_params` `user_index_params` `is_auto_index` `min_index_version` `max_index_version` |

`createIndexes` looped over the index infos and broadcast each one unconditionally — written assuming the etcd half is always there, with no check. So a backup taken without the flag broadcasts an index on **FieldID 0** (`RowIDField`, which is never indexed), and the target ends up with an index entry that gates every future import of that collection:

- the import job finishes importing and then stays in `IndexBuilding`; the client polls `[state=Importing] [progress=80]` and never stops, because `waitBulkInsertState` only returns on `Completed` or `Failed`;
- DataCoord index tasks fail with `parse magic number failed, expected: 16775868, actual: 827474256` — `PAR1`, the index node reading the storage-v2 packed column file of field 0 as a v1 binlog;
- `DescribeIndex` on the restored collection fails with `failed to get collection field: 0`, so the bogus entry is hard to even inspect.

## Change

Whether a backup can serve a secondary restore at all is settled when the backup is created, not at restore time. So check it once, in `Task.Execute`, before any client is created or any DDL is broadcast:

| backup | behaviour |
|---|---|
| every index carries the extra attributes | replay create index as before — unchanged |
| any index does not | fail before anything is restored, naming the indexes and the option to re-create the backup with |

`replayIndex` becomes `checkIndexExtra` and returns only an error. The bool it used to thread through `Task` and `collDDLTask` is gone with it, and so is the warn that announced which indexes were being left out — there is nothing to leave out any more. Net −22 lines.

## Why not restore without the indexes

The first revision degraded instead: no extra attributes meant restoring the collections and their data with no indexes, saying so up front. @huanghaoyuanhhy's review is right that this does not work, and the reasons are now in the doc string:

**A collection with no index cannot be loaded.** `loadCollectionTask.PreExecute` fails with `ErrIndexNotFound` when `DescribeIndex` returns nothing. A loaded source is the ordinary DR case, so the degrade path just moves the failure to the `AlterLoadConfig` broadcast, with a less obvious error. For an unloaded source it yields a secondary that can never be loaded and cannot serve after a failover. Reporting success on a silently diverged secondary is worse than failing — the failure is visible, the divergence is not. This also answers the open question the first revision left: it does not load.

**The backups the degrade path was for barely exist.** Building a secondary needs the backup taken around a flush all, with CDC reading the WAL from its checkpoint; once the primary's WAL is GC'd past that point no secondary can be built from it, indexes or not. The workflow is always made-to-order, so failing on the missing option catches the misconfiguration while it can still be fixed.

**There is precedent in this package.** `checkDynamicField` already fails the restore for the dynamic-field gap behind the same option. This is the same shape and is now handled the same way.

Nothing is reconstructed on the client side. The missing attributes are what makes the DDL replay match the source cluster, and `create_time` in particular does not exist anywhere in a default backup; guessing them would silently put an index with different parameters on the replica.

## Seen in production

A default REST create (`{"async": true, "backup_name": "..."}`, no `with_index_extra`) against a 66-collection cluster this week. `restore secondary` walked past sixty of them before failing on `"test_storage" missing dynamic field in backup` — the dynamic-field check catching it first, behind the same option. Had that check not been there, this path would have been next, and it would have failed later and less clearly.

## Tests

`TestCheckIndexExtra` covers both outcomes plus the partially-populated case, empty index list, empty collection list, and `index_params` missing on its own; `TestIndexNames` covers the message formatting.

`restore secondary --help` now says a backup without the flag is refused, rather than restored without indexes.

## Not in scope

- **Recovering an existing backup.** The attributes can still be read from the source cluster's etcd, so a command that backfills them into an already-taken backup would make those backups restorable. Restore-time is too late for it: `restore secondary` is configured against the target, and `--source_cluster_id` is an identifier, not a connection — this belongs on the source side, as its own command.
- **Making the backup side say so.** `--backup_index_extra` defaults to false, and it also gates the dynamic-field backup through the etcd client created only for it. #1190 covers the visibility half; the contract half — refusing at create time once the user has declared secondary intent — is #1170.

### DIFF
--- a/cmd/restore/secondary.go
+++ b/cmd/restore/secondary.go
@@ -109,8 +109,9 @@ func newSecondaryCmd(opt *root.Options) *cobra.Command {
 		Long: "Restore a backup to a secondary cluster.\n\n" +
 			"Create the backup with --backup_index_extra to restore its indexes too: secondary\n" +
 			"restore replays the source cluster DDL verbatim and needs the index attributes that\n" +
-			"only that flag reads from etcd. A backup created without it restores the collections\n" +
-			"and the data, but without their indexes.",
+			"only that flag reads from etcd. A backup created without it is refused: a collection\n" +
+			"without its indexes cannot be loaded, so restoring one would produce a secondary\n" +
+			"that can never serve after a failover.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			params := opt.InitGlobalVars()
 
--- a/cmd/restore/secondary.go
+++ b/cmd/restore/secondary.go
@@ -106,6 +106,11 @@ func newSecondaryCmd(opt *root.Options) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "secondary",
 		Short: "restore a backup to a secondary cluster",
+		Long: "Restore a backup to a secondary cluster.\n\n" +
+			"Create the backup with --backup_index_extra to restore its indexes too: secondary\n" +
+			"restore replays the source cluster DDL verbatim and needs the index attributes that\n" +
+			"only that flag reads from etcd. A backup created without it restores the collections\n" +
+			"and the data, but without their indexes.",
 		RunE: func(cmd *cobra.Command, args []string) error {
 			params := opt.InitGlobalVars()
 
--- a/core/restore/secondary/coll_ddl_task.go
+++ b/core/restore/secondary/coll_ddl_task.go
@@ -27,8 +27,6 @@ type collDDLTask struct {
 	dbBackup   *backuppb.DatabaseBackupInfo
 	collBackup *backuppb.CollectionBackupInfo
 
-	replayIndex bool
-
 	streamCli milvus.Stream
 	logger    *zap.Logger
 }
@@ -37,10 +35,6 @@ type ddlTaskArgs struct {
 	TaskID string
 
 	BackupInfo *backuppb.BackupInfo
-
-	// ReplayIndex reports whether the backup carries the index attributes that
-	// a verbatim create index replay needs. See replayIndex.
-	ReplayIndex bool
 
 	StreamCli milvus.Stream
 }
@@ -54,8 +48,6 @@ func newCollDDLTask(args ddlTaskArgs, dbBackup *backuppb.DatabaseBackupInfo, col
 		backupInfo: args.BackupInfo,
 		dbBackup:   dbBackup,
 		collBackup: collBackup,
-
-		replayIndex: args.ReplayIndex,
 
 		streamCli: args.StreamCli,
 		logger:    log.With(zap.String("task_id", args.TaskID), zap.String("ns", ns.String())),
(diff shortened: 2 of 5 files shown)


huanghaoyuanhhy:
**Suggest failing the "none has extra" branch as well, rather than skipping indexes:**

1. **The degrade path does not actually work.** Answering the open question: a collection without indexes cannot be loaded — `loadCollectionTask.PreExecute` fails with `ErrIndexNotFound` when `DescribeIndex` finds no index (milvus `internal/proxy/task.go`). A loaded source is the common DR case, so skipping indexes just fails later, at the `AlterLoadConfig` broadcast, with a less obvious error. And even for unloaded sources it yields a secondary that can never be loaded without a manual `CreateIndex` and cannot serve after failover. Reporting success on a silently diverged secondary is worse than failing — the failure is visible, the divergence is not.
2. **The backups this path is meant to rescue barely exist.** Building a secondary requires the backup to be taken around a flush all and CDC to read the WAL from the backup checkpoint; once the primary's WAL is GC'd past that point, no secondary can be built from it regardless of indexes. Backups taken before the `--backup_index_extra` contract was known have almost certainly lost that window. The realistic workflow is always made-to-order: take a fresh backup with the flag when building a secondary — and failing fast on a missing flag catches the misconfiguration at exactly the right moment.
3. **Precedent in this package:** `checkDynamicField` already fails the restore for the dynamic-field gap gated by the same flag. The index gap is the same shape; the handling should be the same.

So `replayIndex` should return an error when indexes exist but none carries the extra info, naming the indexes and pointing at `--backup_index_extra`, same as the inconsistent case.

czs007:
Agreed on all three, and the third one played out in production this week.

`replayIndex` is now `checkIndexExtra` and returns only an error: any index missing the etcd attributes fails the restore, whether that is all of them or some of them. The bool it used to thread through `Task` and `collDDLTask` is gone, along with the warn that announced which indexes were being left out — there is nothing to leave out any more. Net −22 lines.

On the point about `checkDynamicField` being the precedent: a customer hit exactly that this week. A default REST create with no `with_index_extra`, 66 collections, and `restore secondary` walked past sixty of them before failing on `"test_storage" missing dynamic field in backup`. Same option, same silent gap, and the error pointed at the backup rather than the option — which is what #1190 addresses on the visibility side. Had the dynamic-field check not been there, the index path would have been next, and it would have failed later and less clearly.

Also took the second point into the doc string, since it is the part that is easy to lose: the degrade path is not just worse than failing, the backups it was written for cannot build a secondary anyway once the WAL has moved past their checkpoint.

`cmd/restore/secondary.go` help text updated to match — a backup without the flag is refused now, not restored without indexes.

gofmt, build and golangci-lint clean; `./core/restore/...` and `./cmd/...` pass.

Issue #1176, opened 2026-08-20
[Bug]: a second restore secondary is silently discarded and the command hangs forever


czs007:
### Current Behavior

A second `restore secondary` into a secondary that has already received one is
silently discarded, and the command then hangs forever instead of failing.

This is not an exotic case: a `restore secondary` that fails partway leaves the
target advanced, so **retrying it does nothing** — the retry hangs, and the only
way forward is to reset the secondary (force promote it to a standalone primary
and configure it as a secondary again).

Two defects combine.

**1. The wire time ticks restart at 1 on every run.**

`StreamClient.ttCounter` (`internal/client/milvus/stream.go`) is a plain
`uint64` field with no seed, and every `restore secondary` process builds a
fresh `StreamClient`:

```go
type StreamClient struct {
	...
	ttCounter  uint64 // guarded by dispatchMu
}

func (s *StreamClient) Send(ctx context.Context, build func(ts uint64) []message.MutableMessage) error {
	s.ttCounter++
	bodyTS := s.ttCounter
	...
	imm := msg.WithTimeTick(wireTS).WithLastConfirmed(NewFakeMessageID(wireTS))...
```

So every restore stamps its injected messages `1, 2, 3, ...`. Two consecutive
restores of different backups both log
`base:{msg_type:CreateCollection timestamp:2}`.

The secondary only ever moves its replicate checkpoint forward and drops
anything at or below it
(`internal/streamingnode/server/wal/interceptors/replicate/replicates/impl.go`):

```go
checkpoint := impl.secondaryState.GetCheckpoint()
if (isTxnBody && rh.TimeTick < checkpoint.TimeTick) || (!isTxnBody && rh.TimeTick <= checkpoint.TimeTick) {
	return nil, status.NewIgnoreOperation("message is too old, message_id: %s, time_tick: %d, txn: %t, current time tick: %d", ...)
}
```

After a first restore the checkpoint holds a real TSO, so the second restore's
ticks are all far below it. Observed on the target:

```
message is too old, message_id: 1, time_tick: 1, txn: false, current time tick: 468520367335407637
message is too old, message_id: 2, time_tick: 2, txn: false, current time tick: 468520367335407648
message is too old, message_id: 3, time_tick: 3, txn: false, current time tick: 468520367335407637
```

`IgnoreOperation` is not an error to the sender, so nothing is reported: the
DDL, the index and the import message are all dropped and the tool proceeds as
if they had been accepted.

[...]

PR #1177, opened 2026-08-20
restore: stop waiting for an import job the target does not have

czs007:
issue: #1176

A secondary restore ships the import job id inside the import message and then
polls the target for that job:

```go
jobID := rand.Int64()
...
body := &message.ImportMsg{ ..., JobID: jobID }
```

If the message was not applied, the target has no such job. `GetBulkInsertState`
still succeeds and reports an empty state, which is neither `Failed` nor a state
`waitBulkInsertState` is waiting for, so the loop polls every three seconds
forever. The one guard in the loop does not end it either — after an hour with
no progress it logs a warning and resets its own timer.

The state is easy to reach. `StreamClient.ttCounter` starts at zero in every
process, so every restore time-ticks its injected messages from 1, while a
secondary only moves its replicate checkpoint forward and discards anything at
or below it:

```
message is too old, message_id: 1, time_tick: 1, txn: false, current time tick: 468520367335407637
message is too old, message_id: 2, time_tick: 2, txn: false, current time tick: 468520367335407648
```

That is an ignore, not an error, so nothing is reported to the sender. A restore
that failed part way does not roll back, so retrying it is enough to get here:
the DDL, the index and the import message are all dropped and the command hangs
with nothing written and nothing logged.

This does not change how time ticks are assigned — the small ticks are
load-bearing, since `Forward` replays the backup's flush-all messages with the
source's original time ticks and the injected ones have to stay below them. It
makes the unknown-job case observable instead: an empty state that persists for
a minute ends the wait with an error that names the job and describes the reset
the target needs. Only leaving the secondary role drops the replicate
checkpoint, so re-applying the same replicate configuration does not clear it
and neither does restarting the target; that is worth saying in the error,
because it is the first thing an operator tries.

A job that is merely slow to appear is unaffected: the window only starts when
an empty state is seen, and it restarts if the job becomes known again.

### Reproduction

Two collections, two backups, two restores into the same secondary, no reset in
between:

| | first restore | second restore |
|---|---|---|
| injected `CreateCollection timestamp` | 2 | 2 |
| outcome | `restore done` | hangs; killed after 200s |
| bulk insert polls | 5 | 66, every one with an empty state |
| collection on the target | present, 200 rows | never created |

After resetting the target (force-promote to standalone primary, then configure
it as a secondary again) the identical backup restored on the first attempt.

### Tests

`TestWaitBulkInsertStateUnknownJob` covers the four cases: an always-unknown job
gives up and says how to recover, a job that is slow to appear is still waited
for, the window restarts once the job becomes known again, and a failed job
still reports its own reason.

### DIFF
--- a/core/restore/secondary/coll_dml_task.go
+++ b/core/restore/secondary/coll_dml_task.go
@@ -49,6 +49,10 @@ var (
 	// target only knows the job once that message has been applied; the window
 	// covers the message still being in flight.
 	_bulkInsertUnknownJobTimeout = 1 * time.Minute
+	// How long to let the target apply the DDL before concluding that a
+	// collection the restore created is not there.
+	_restoreVerifyTimeout  = 30 * time.Second
+	_restoreVerifyInterval = 3 * time.Second
 )
 
 type partitionDir struct {
@@ -380,15 +384,16 @@ func (dmlt *collDMLTask) waitBulkInsertState(
 				unknownSince = time.Now()
 			} else if time.Since(unknownSince) >= _bulkInsertUnknownJobTimeout {
 				return "", fmt.Errorf("secondary: the target does not know import job %d after %s, "+
-					"so the import message this restore sent was never applied. The usual reason is "+
-					"that the target is not a freshly configured secondary: a secondary only moves "+
-					"its replicate checkpoint forward, and the messages a restore injects are "+
-					"time-ticked from 1, so everything this restore sends is discarded as too old "+
-					"once an earlier restore has advanced that checkpoint. Reset the target before "+
-					"restoring again: force-promote it to a standalone primary, which is the only "+
-					"transition that drops the checkpoint (re-applying the same replicate "+
-					"configuration does not, and neither does restarting it), drop whatever an "+
-					"earlier restore left behind, then configure it as a secondary again",
+					"so the import message this restore sent was never applied. A secondary restore "+
+					"targets a newly deployed secondary and runs once: the messages it injects are "+
+					"time-ticked from 1, while a secondary only ever moves its replicate checkpoint "+
+					"forward, so a target that has already been restored discards everything a "+
+					"second restore sends. There is no way to return a used target to that state -- "+
+					"re-applying the same replicate configuration does not clear the checkpoint, "+
+					"restarting the target does not, and dropping its collections makes matters "+
+					"worse, because their ids stay reserved while the collections are being "+
+					"reclaimed and a later restore of the same ids is then discarded without an "+
+					"error. To bootstrap again, deploy a new secondary and restore into that",
 					jobID, _bulkInsertUnknownJobTimeout)
 			}
 		} else {
--- a/core/restore/secondary/coll_dml_task.go
+++ b/core/restore/secondary/coll_dml_task.go
@@ -36,9 +36,19 @@ import (
 const (
 	_bulkInsertTimeout             = 60 * time.Minute
 	_bulkInsertRestfulAPIChunkSize = 256
-	_bulkInsertCheckInterval       = 3 * time.Second
 	_commitImportMessageType       = 45
 	_commitImportMessageVersion    = 2
+)
+
+// Overridden in tests so the unknown-job path does not take a minute to
+// exercise.
+var (
+	_bulkInsertCheckInterval = 3 * time.Second
+	// How long an import job may stay unknown to the target before the restore
+	// gives up on it. The job id travels inside the import message, so the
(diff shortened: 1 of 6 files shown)


czs007:
Updated after establishing what the contract actually is, and after finding that
the recovery this PR originally described does not work.

**A secondary restore targets a newly deployed secondary and runs once.** That
was never written down and never checked, and every way of violating it ends in
silence:

| violation | what happens today |
|---|---|
| the target already holds the collection | the create is skipped as a duplicate and the data is imported into the existing collection, doubling its rows |
| the target was already restored | every injected message is discarded as too old; the restore then waits forever on an import job that was never created |
| the target's collections were dropped first | their ids stay reserved while they are reclaimed, so the replayed create is skipped without an error and the restore reports success against an empty target |

The third one is the reason this PR changed. Its original error message told the
operator to "drop whatever an earlier restore left behind", which walks straight
into it. Verified on a two-cluster pair: after dropping the target's collection
and restoring the same backup, the restore reports `restore done` with exit code
0 and the target has no collection at all. The target logs
`collection already created, skip add collection to meta table` — rootcoord's
idempotency guard in `AddCollection` matches on collection id alone, and the id
of the dropped collection is still held in `CollectionDropping` state until
`GcConfirm` reports the collection's segment metadata gone. Restoring a backup
of a collection the target has never held succeeds at the same moment, on the
same target, with the same replicate checkpoint — the only difference being the
id.

### What this adds

- `checkTargetIsUnused` refuses before anything is injected when the target
  already holds one of the collections in the backup, naming them.
- `verifyRestored` confirms the collections are actually present before
  reporting success. Every other step reports on whether the messages it sent
  were accepted, not on what the target did with them, which is how a restore
  could report success having created nothing.
- The unknown-import-job wait still ends after a minute rather than polling
  forever; its message now states the contract instead of prescribing a recovery
  that does not work.

### Tests

`TestCheckTargetIsUnused` — a clean target is accepted; a target holding one of
the collections is refused and it is named; a database that does not exist yet
is not a reason to refuse.

`TestVerifyRestored` — present collections pass; a collection that never appears
fails with the contract in the message; one that appears a moment later still
passes; a failing lookup is reported rather than treated as missing.

`TestWaitBulkInsertStateUnknownJob` — unchanged in substance, message assertion
updated.

huanghaoyuanhhy (commented):
Non-blocking, just thinking out loud.

The 1-minute window is a heuristic over the in-flight latency of the import message rather than something derived from a guarantee — if a loaded target legitimately takes longer than a minute to apply the message, this fires spuriously, and when the message was discarded in the first second the restore still burns the full minute before reporting.

A deterministic signal seems available: the injected import message carries a known time tick, and `checkTargetNotRestored` already reads the per-pchannel replicate checkpoint via `GetReplicateInfo`. Once the target's checkpoint has moved past the message's time tick and the job is still unknown, the message has been consumed without creating the job and no amount of further waiting will help — that could be the give-up condition, with the timeout kept only as a safety net (or dropped).

I suspect the job→pchannel/tick bookkeeping through the txn message path is what makes this annoying, and the two upfront checks already catch the common used-target cases, so fine to leave as is for now.

core/restore/secondary/coll_dml_task.go
```diff
@@ -36,11 +36,25 @@ import (
...
 	_bulkInsertTimeout             = 60 * time.Minute
 	_bulkInsertRestfulAPIChunkSize = 256
-	_bulkInsertCheckInterval       = 3 * time.Second
 	_commitImportMessageType       = 45
 	_commitImportMessageVersion    = 2
 )
 
+// Overridden in tests so the unknown-job path does not take a minute to
+// exercise.
+var (
```
huanghaoyuanhhy, line 45: Non-blocking, just thinking out loud.

The 1-minute window is a heuristic over the in-flight latency of the import message rather than something derived from a guarantee — if a loaded target legitimately takes longer than a minute to apply the message, this fires spuriously, and when the message was discarded in the first second the restore still burns the full minute before reporting.

A deterministic signal seems available: the injected import message carries a known time tick, and `checkTargetNotRestored` already reads the per-pchannel replicate checkpoint via `GetReplicateInfo`. Once the target's checkpoint has moved past the message's time tick and the job is still unknown, the message has been consumed without creating the job and no amount of further waiting will help — that could be the give-up condition, with the timeout kept only as a safety net (or dropped).

I suspect the job→pchannel/tick bookkeeping through the txn message path is what makes this annoying, and the two upfront checks already catch the common used-target cases, so fine to leave as is for now.

PR #1179, opened 2026-08-21
storage: name the account-qualified host in azure snapshot uris

### LINKED ISSUE #1178: [Bug]: Snapshot-format backup on Azure fails: Milvus rejects URI endpoint "blob.core.windows.net:443"
**Milvus-Backup Version**

main @ 4b5b3ee (built 2026-08-21)

**Can you reproduce this issue on the latest version?**

Yes, it also happens on the latest version — the Azure snapshot path was added in #1140 and is unchanged on main.

**Current Behavior**

On an Azure-backed Milvus master cluster, a snapshot-format backup fails a couple of seconds after creation with `BACKUP_FAIL`. The backup task log shows:

```
backup: export snapshot: client: export snapshot failed: client: operation failed: error_code:IllegalArgument reason:"Azure snapshot URI endpoint \"blob.core.windows.net:443\" does not match the configured endpoint: invalid parameter"
```

The task logs the target path as `azure://blob.core.windows.net:443/<bucket>/<backup-root>/<backup-name>/bundle` with `external_spec_set=true`, then the export is rejected by Milvus.

**Expected Behavior**

The ExportSnapshot request passes Milvus's snapshot URI validation and the export proceeds.

**Steps To Reproduce**

1. Run a Milvus master cluster backed by Azure Blob Storage (endpoint `core.windows.net:443`).
2. Create a backup that resolves to the snapshot format, passing the Azure storage spec explicitly (the Zilliz Cloud control plane always passes one, so `external_spec_set=true`; any caller setting `storage spec` fields on the create request hits the same path).
3. The task fails at `export snapshot` with the IllegalArgument error above.

**Environment**

- milvus-backup: main @ 4b5b3ee
- Milvus: master-latest (snapshot export is master-only; 2.6 uses the binlog format and is unaffected)
- Object storage: Azure Blob, westus3, workload-identity (IAM) credentials

**Anything else?**

Root cause, verified against both codebases. milvus-backup builds the Azure snapshot URI as `azure://blob.<service-endpoint>/<bucket>/<key>` (internal/storage/snapshot.go, `SnapshotURI`): the host is the service endpoint with a literal `blob.` prefix and no storage account name. Milvus's snapshot validator (internal/snapshotio/storage/validator.go, `applySnapshotURILocationToConfig` + `inferStorageEndpointIdentity`) only accepts an Azure URI endpoint that is (a) account-qualified (`<account>.blob.<suffix>`), (b) the bare suffix (`core.windows.net`), or (c) exactly equal to the instance's effective endpoint `<account>.blob.<configured address>`. The `blob.<suffix>` form matches none of these, so the export is rejected whenever `external_spec` is set. The Milvus design doc (20260609-external-snapshot-export-restore.md) also specifies `azure://<account-endpoint>/<container>/<object-key>`. #1164's endpoint omission deliberately does not cover Azure (there is no endpoint-less Azure URI form), so this path always embeds the endpoint and always fails validation.

Suggested fix direction: emit the bare-suffix form `azure://core.windows.net:443/<bucket>/<key>` (drop the `blob.` prefix). Traced through Milvus's validator and resolver, this form is accepted and still resolves to the same-backend instance config.

huanghaoyuanhhy:
## Summary

Fixes #1178

Snapshot-format backup on Azure failed Milvus's snapshot URI validation because milvus-backup built the URI host as `blob.<service-endpoint>` with no storage account name — a form the validator never accepts (it recognizes `<account>.blob.<suffix>`, the bare `<suffix>`, or an exact match with the instance's effective endpoint). The URI now names the account-qualified endpoint, matching the Milvus snapshot URI contract (`azure://<account-endpoint>/<container>/<key>`).

## Changes

- storage: build Azure snapshot URIs as `azure://<account>.blob.<service-endpoint>/<bucket>/<key>` from the configured account name, and fail fast when the account name is unset
- storage: drop the default `:443` port from the URI so it compares equal to the metadata URI Milvus reports back for the completed export
- storage: treat a `milvusAddress` override as the full host Milvus resolves (account prefix included), used verbatim
- test: cover the account-qualified form, the missing-account-name error, and the Milvus-view override

Note: with this fix, an Azure deployment whose backup container lives in a different storage account than the Milvus data (the uat3 topology in #1178) still fails at Milvus's resolver, which only supports same-account server-side copies on Azure. That is a Milvus-side limitation to track separately.

### DIFF
--- a/internal/storage/snapshot.go
+++ b/internal/storage/snapshot.go
@@ -32,17 +32,33 @@ func SnapshotURI(cfg Config, key string) (string, error) {
 		return "", err
 	}
 
-	// Azure names its service endpoint in the host and carries the account name in the
-	// extfs access_key_id, so the host is the blob. service endpoint, not the account one.
+	// Milvus's snapshot validator recognizes an Azure URI host only as the
+	// account-qualified endpoint (<account>.blob.<service-endpoint>) or the bare
+	// service endpoint, and rejects anything else as not matching the configured
+	// endpoint — so the URI names the account-qualified host. The account name is
+	// always set, since the Azure client itself needs it to build the blob
+	// service URL.
 	if cfg.Provider == v2.ProviderAzure {
-		host := endpointHost(cfg.Endpoint)
 		if cfg.MilvusEndpoint != "" {
-			host = endpointHost(cfg.MilvusEndpoint)
+			// The override names the exact host Milvus resolves, account included.
+			host := endpointHost(cfg.MilvusEndpoint)
+			if host == "" {
+				return "", fmt.Errorf("storage: snapshot uri for azure needs an endpoint")
+			}
+			return fmt.Sprintf("azure://%s/%s/%s", host, cfg.Bucket, key), nil
 		}
+		if cfg.Credential.AzureAccountName == "" {
+			return "", fmt.Errorf("storage: snapshot uri for azure needs an account name")
+		}
+		host := endpointHost(cfg.Endpoint)
 		if host == "" {
 			return "", fmt.Errorf("storage: snapshot uri for azure needs an endpoint")
 		}
-		return fmt.Sprintf("azure://blob.%s/%s/%s", host, cfg.Bucket, key), nil
+		// Milvus reports the completed export's metadata URI with the default
+		// https port dropped; drop it here too, or the backup meta's cross-check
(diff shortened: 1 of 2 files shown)


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

PR #1181, opened 2026-08-23
storage: send source sas for cross-account azure snapshots

### LINKED ISSUE #1180: [Feature]: support cross-account Azure snapshot backup/restore via extfs source_sas_token
### Is your feature request related to a problem? Please describe.

Snapshot-format backup fails whenever the Milvus instance bucket and the backup bucket live in **different Azure storage accounts** — the default topology for a dedicated backup/DR account:

```
[ERROR] [server/create.go:297] ["create backup task execute fail"] [error="backup: run collection task: ... client: export snapshot failed: client: operation failed: error_code:IllegalArgument reason:\"server-side cross-bucket copy is unsupported for this provider/endpoint pair; cross-service copy would require streaming (unsupported): invalid parameter\" code:1100 ..."]
```

What happens today:

- milvus-backup auto-selects the snapshot format when the server supports it (`core/backup/task.go`, `resolveFormat`), and builds the export target + extfs in `core/backup/snapshot_target.go` via `storage.SnapshotStoreURI` / `storage.SnapshotExternalSpec` (`internal/storage/snapshot.go`).
- For cross-backend targets, `newSnapshotTarget` attaches the backup storage's AK/SK as the extfs credential — but on Azure a source blob in another storage account **cannot be read by any single-principal credential**: per the [Copy Blob From URL authorization table], neither account's Shared Key nor the request's own Entra ID token authorizes a cross-account source read; only a read-scoped SAS on the source URL does.
- Milvus therefore rejects the request up front (`internal/snapshotio/storage/resolver.go`, `validateProviderEndpointPair` requires `sameAzureAccountEndpoint`). Note `internal/storage/factory.go:SameBackend` already treats a different `AzureAccountName` as a different backend — milvus-backup knows this topology is cross-account, but has nothing to send for it.

Current workarounds: put the backup container in the same storage account as Milvus, or fall back to `--format binlog` (streamed through milvus-backup, losing the server-side-copy benefit).

### Describe the solution you'd like.

Milvus has just added support for this: `external_spec.extfs.source_sas_token`, an Azure-only, read-scoped SAS for the copy source (milvus-io/milvus#52770, design doc §10 of `20260609-external-snapshot-export-restore.md`). Track that work here and use it:

1. **Backup (export)** — the copy source is the Milvus instance bucket. When `SameBackend(milvusCfg, backupCfg)` is false **and both providers are Azure**, attach `source_sas_token` to the extfs built by `storage.SnapshotExternalSpec` (`core/backup/snapshot_target.go`, `core/backup/coll_snapshot_task.go`). The SAS must be readable on the Milvus instance container; milvus-backup already holds the instance account key in its config, so it can mint a service SAS itself, or accept an operator-provided one.
2. **Restore** — the copy source is the backup bucket. `core/restore/snapshot_source.go` builds the extfs from the backup storage config and needs the same treatment, with the SAS minted from (or provided for) the backup account.
3. **Config surface** — something like `source_sas_token` (explicit, pre-minted) and/or `sas_ttl` (auto-mint a container-scoped read SAS at task start) under the azure storage sections. Auto-minting needs care: shared-key configs can only sign service/account SAS (a user-delegation SAS requires Entra credentials), and the TTL must cover the whole export/restore job window since Milvus retains the spec until the job is terminal.
4. Validation should stay fail-closed on our side too: SAS only for azure→azure cross-account snapshot tasks, never for same-account (Milvus rejects that as an input error) or for the binlog format.

### Describe an alternate solution.

Operator-only path with no new code: keep using `--format binlog` for cross-account Azure. It works today but streams all data through milvus-backup and forgoes the snapshot consistency semantics — that is exactly what this feature exists to avoid.

### Anything else? (Additional Context)

Upstream references:

- Milvus feature issue: milvus-io/milvus#52769
- Milvus implementation PR (adds `extfs.source_sas_token`, endpoint-pair relaxation, source-URL SAS): milvus-io/milvus#52770
- Design doc: `docs/design-docs/design_docs/20260609-external-snapshot-export-restore.md` §10 in the milvus repo
- Azure copy authorization rules: https://learn.microsoft.com/en-us/rest/api/storageservices/copy-blob-from-url

Dependency note: milvus-backup does **not** gate sending `source_sas_token` on server capability or version detection. The field rides the snapshot feature itself: a server that supports snapshots but does not know the extfs key rejects the request outright and the task fails with that error, which is the accepted behavior — no probe, no flag, the failure points at the mismatch directly. milvus-io/milvus#52770 landing in a release milvus-backup targets remains a prerequisite for the copy to actually succeed.

huanghaoyuanhhy:
Fixes #1180

## What

Snapshot-format backup and restore now work when the Milvus instance bucket
and the backup container live in **different Azure storage accounts**: the
extfs handed to Milvus carries `source_sas_token`, the read-scoped SAS that
authorizes the source side of the copy (milvus-io/milvus#52770 contract).

- **Backup** (`core/backup/snapshot_target.go`): the copy source is the
  instance bucket, so the SAS is resolved from the milvus storage side.
- **Restore** (`core/restore/snapshot_source.go`): the copy source is the
  backup bucket, so the SAS is resolved from the backup storage side.
- **Minting** (`internal/storage/sas.go`): an explicit token wins; otherwise a
  container-scoped Read+List SAS is minted per task — signed locally with the
  shared key, or via a user delegation key for IAM credentials — valid for 48h,
  the same budget the streamed-copy SAS uses, since Milvus keeps the spec until
  the job goes terminal.
- **Config**: `milvus.storage.sourceSASToken` / `backup.storage.sourceSASToken`
  (secret, azure-only, env `*_STORAGE_SOURCE_SAS_TOKEN`). The field does not
  inherit across sections: a SAS is scoped to one account.

## Fail-closed rules

`CrossAccountAzure` compares storage **accounts**, not endpoints — the same
account behind a private link is not a crossing. A SAS is attached only for
azure→azure cross-account snapshot copies; a token configured for anything else
(same account, non-azure end, binlog topology) is rejected at task construction
with the rule named, matching the server's own input validation.

## No capability gating

Per the decision recorded on the issue, sending the SAS is **not** gated on
server feature detection: the field rides the snapshot feature itself, and a
server that supports snapshots but rejects the unknown extfs key fails the task
outright with that error. milvus-io/milvus#52770 landing in a release we target
remains the prerequisite for the copy to succeed.

## Tests

Unit tests cover the mint shapes (explicit passthrough incl. `?` trimming,
shared-key service SAS query fields, IAM needs shared key or delegation),
`CrossAccountAzure` (incl. same-account-two-endpoints), the extfs rendering and
its non-azure rejection, the backup/restore constructors (mint, explicit,
same-account and non-azure fail-closed), plus config validation and render
skipping. `go test ./...` and `golangci-lint` are clean.

### DIFF
--- a/cmd/restore/restore.go
+++ b/cmd/restore/restore.go
@@ -205,7 +205,7 @@ func (o *options) run(cmd *cobra.Command, params *v2.Config) error {
 		return err
 	}
 
-	task, err := restore.NewTask(args)
+	task, err := restore.NewTask(cmd.Context(), args)
 	if err != nil {
 		return err
 	}
--- a/configs/backup-azure.yaml
+++ b/configs/backup-azure.yaml
@@ -23,6 +23,12 @@ milvus:
     # The Azure storage account, required by the azure provider. Prefer
     # MILVUS_STORAGE_ACCOUNT_NAME over writing it here.
     accountName: "<your-storage-account>"
+    # A read-scoped SAS for this account's container, for snapshot-format
+    # backups kept in a DIFFERENT storage account: Azure lets no single
+    # credential read across accounts, so Milvus copies under this token.
+    # Leave it out to have milvus-backup mint one from the account key for
+    # the duration of each task.
+    # sourceSASToken: "<your-source-sas>"
     # For azure the bucket is the blob container.
     bucketName: "milvus-data"
     rootPath: "milvus/azure-instance"
--- a/core/backup/snapshot_target.go
+++ b/core/backup/snapshot_target.go
@@ -1,6 +1,7 @@
 package backup
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/url"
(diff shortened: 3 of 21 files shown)


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

Issue #1171, opened 2026-08-20
[Enhancement]: Transport-neutral usecase layer as the foundation for HTTP API v2


huanghaoyuanhhy:
## What would you like to be added?

A usecase layer in `core/app`: one struct per action, dependencies injected through the constructor, transport-neutral in and out. This is the foundation #1124 (HTTP API v2) mounts on, and its first step.

```go
// core/app — each usecase is its own struct; there is no shared god-object.
type GetBackup struct {
	cli      storage.Client
	taskMgr  *taskmgr.Mgr
	rootPath string
}

// BackupView keeps the two resource halves separate. Merging them is a
// rendering decision of the transport, not a property of the action.
type BackupView struct {
	Task     taskmgr.BackupTaskView // job half; nil when the task is unknown (e.g. after a restart)
	Meta     *backuppb.BackupInfo   // artifact half; nil while the job is still in flight
	MetaSize int64
}

func (uc *GetBackup) Execute(ctx context.Context, req GetBackupRequest) (*BackupView, error)
```

Design rules:

- **One struct per usecase, no manager object.** Each usecase declares exactly the dependencies it needs — `GetRestore` takes only the task manager and never touches storage, `ListBackups` takes only the backup storage client. Construction is eager and explicit; there is no lazy client creation, no locking, no lifecycle.
- **Every action lives in the layer, no exceptions.** Thin actions like `list` and `get_restore` are usecases too, even though the extraction saves almost no lines by itself. The point is that transports never import `internal/meta`, `internal/storage`, `internal/taskmgr` or `pbconv` directly: `cmd` and `core/server` talk to `core/app` and nothing below it, so the dependency tree stays uniform and each usecase remains independently replaceable.
- **Transports own their input and output contracts.** Flag parsing, query params, deprecated pb fields (`force`, `db_collections`, ...) and validation-to-error-code mapping stay in `cmd`/`core/server`. Rendering stays there too: the v1 handler merges a `BackupView` through `pbconv.NewBackupInfoBrief`, the future v2 handlers render the artifact half as `backups/describe` and the job half as `jobs/backup/describe`, and the CLI prints. A usecase never sees a pb request and never returns the merged v1 shape.
- **Actions are synchronous.** Async execution is a deployment concern of the HTTP server (goroutine + taskmgr registration), not a flag in the request struct.

[...]

PR #1182, opened 2026-08-25
app: extract the list backups usecase

huanghaoyuanhhy:
## Summary
First usecase of the `core/app` layer proposed in #1171, extracted from the thinnest action: `list`. Both transports — the `/api/v1/list` handler and the `list` CLI command — now call `app.ListBackups` instead of assembling `storage.NewBackupStorage` + `meta.List` themselves, and no longer import `internal/meta` / `internal/storage`.

## Changes
- add `core/app` with the `ListBackups` usecase: one struct, dependencies injected through the constructor, transport-neutral in and out; backup-storage client construction moves behind the layer
- rewire `core/server/list.go` and `cmd/list/list.go` onto the usecase; swagger annotations, the deprecated `collection_name` rejection and the v1 wire shape stay in the transport unchanged
- cover the usecase with mocked-storage tests: summaries per backup, skip-unreadable-meta, root-listing failure

Behavior is unchanged: the storage client is still created per request.

Related to #1171


lentitude2tk (approved):
/LGTM

PR #1188, opened 2026-08-27
restore: carry struct-level nullable through schema conversion

huanghaoyuanhhy:
## Background

Milvus keeps nullable authoritative at the struct level of an array-of-struct field: it is rejected when set per sub-field and instead propagated onto every sub-field by proxy before persisting. Because `DescribeCollection` returns those propagated child flags verbatim while struct-level handling depends on every consumer keeping `StructArrayFieldSchema` intact, any conversion that models structs as flat field lists loses the parent bit and re-created collections fail validation (`sub-field in non-nullable struct cannot be nullable individually, ...`).

The asymmetry itself is tracked upstream at https://github.com/milvus-io/milvus/issues/52924 — this PR is deliberately a downstream compatibility pass so backup behaves like the server does, not a statement about where the UX fix belongs.

## Changes

- `core/proto`: add `bool nullable = 5` to `StructArrayFieldSchema`, regenerate generated code
- `core/backup`: carry the struct-level flag verbatim from DescribeCollection
- `core/restore` (primary DDL path + secondary conv path): rebuild it per server rule via shared `conv.ApplyStructNullable`; derive it from nullable children for backup metas written before this field existed
- sync swagger definition

## Tests

Unit tests cover all three meta shapes (struct flag set → propagated; legacy child-only flags → repaired; both unset → untouched) plus the describe wiring. Full build and golangci-lint clean.

Related to https://github.com/milvus-io/milvus/issues/52924


lentitude2tk (approved):
/LGTM

PR #1197, opened 2026-09-02
mpath: document the snapshot bundle layout

huanghaoyuanhhy:
/kind improvement

The header comment in internal/storage/mpath only covered the milvus and binlog-backup layouts. Add the snapshot bundle layout written by Milvus ExportSnapshot — each export is namespaced under `bundle/exports/<uuid>` (datacoord's snapshot_export_manager), with the snapshot metadata and one avro manifest per segment under `snapshots/...`, the data files under `files/` keeping their milvus-root-relative paths, and a transient `_staging/` — verified against milvus's snapshotio/storage and datacoord/snapshot_export_manager.

Correct the existing entries along the way:

- the binlogs prefix was misspelled binglogs
- log_idx is a log ID
- note L0 delta logs carry partition -1
- note the group id level is absent in backups that predate it, which is what restore's fallback handles

Also fix the path parse error reports indexing the wrong capture group — on an id that overflows int64 they panicked instead of returning the error — and cover that with overflow test cases.

### DIFF
--- a/internal/storage/mpath/path.go
+++ b/internal/storage/mpath/path.go
@@ -8,25 +8,47 @@ import (
 	"strings"
 )
 
-// milvus bin log path
+// milvus bin log path (${root} is the milvus storage root)
 // insert log
-// ${root}/insert_log/${collection_id}/${partition_id}/${segment_id}/${field_id}/${log_idx}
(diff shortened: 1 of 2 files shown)


lentitude2tk (approved):
/LGTM

PR #1200, opened 2026-09-03
app: extract the delete backup usecase

huanghaoyuanhhy:
Part of #1171.

## What

`core/app` gains the `DeleteBackup` usecase: one struct holding the backup storage client and the root path, constructor-built from config, with `Execute(ctx, name)` as the transport-neutral action. Both transports call it and `core/del` goes away:

- the `delete` CLI command (`cmd/del`) calls `app.NewDeleteBackup` + `Execute` instead of assembling a storage client and a `core/del.Task` itself
- the `/api/v1/delete` handler (`core/server`) does the same, through a `newDeleteBackup` constructor hook on the server config that tests stub out — the same seam list got in #1182
- `core/del` had no callers left, so the package folds into the usecase

## Semantics unchanged

The meta must be readable before anything is removed, then the whole backup dir prefix is deleted. A name with no readable backup behind it is still an error, not a silent success. The v1 wire shape, request_id defaulting, and error-to-code mapping stay in the handler.

## Tests

- `core/app/delete_test.go`: deletes every object under the backup dir; refuses when the meta is unreadable (nothing deleted); fails when the delete listing fails
- `core/server/del_test.go`: httptest cases over the handler contract — success path, missing name rejected with `Parameter_Error` before the usecase runs, request_id generated and forwarded, constructor and execute errors mapped to `Fail`


lentitude2tk (approved):
/LGTM

lentitude2tk (approved):
/LGTM

PR #1201, opened 2026-09-04
storage: rename azure DeleteObject param prefix to key

huanghaoyuanhhy:
/kind improvement

`AzureClient.DeleteObject` deletes a single object, but its parameter was named `prefix`, unlike the other `Client` implementations (minio / local / gcp native) which all use `key`. Renamed to `key` to match the interface — no behavior change.


lentitude2tk (approved):
/LGTM

PR #1202, opened 2026-09-04
app: move the usecase layer out of core

huanghaoyuanhhy:
/kind improvement

`core/app` — the usecase layer — sat at the same level as the primitives it
composes (`backup`, `restore`, `check`, ...), which muddied what `core/` means.
Move it to a top-level `app/` so the tree reads top-down:

- `cmd/`, `core/server` — transports
- `app` — usecases
- `core/*` — primitives
- `internal/*` — infrastructure

Pure move plus an import rewrite in the five importers (`cmd/list`,
`cmd/del`, `core/server`). No behavior change.


lentitude2tk (approved):
/LGTM

PR #1203, opened 2026-09-04
app: extract the get backup and get backup task usecases

### LINKED ISSUE #890: `/get_backup` API succeeds with empty result for invalid `backup_name`
Hi, I have a question regarding the behavior of the `/get_backup` API when an invalid/non-existent `backup_name` is provided.
When I call `get_backup` with a `backup_name` that does not exist, the API returns **success**, but the returned backup info is always empty.
(milvus-backup version is v0.5.9)
```bash
$ curl -XGET "http://localhost:8080/api/v1/get_backup?backup_name=nonexist" 

{
  "msg": "success",
  "data": {
    "state_code": 0,
    "progress": 0,
    "size": 0,
    "meta_size": 0
  }
```

This seems to be caused by the following code
https://github.com/zilliztech/milvus-backup/blob/8c6e5f878875913f3206c366ce8a6f624aafc9c4/core/server/get_backup.go#L90-L93

If the given `backup_name` does not exist, I expected `get_backup` to return an error (e.g., NotFound).
In contrast, `get backup by id` fails properly when the backup does not exist, so this behavior difference is a bit confusing.

Is returning success with an empty payload for a non-existent `backup_name` an intended behavior?
If it is intended, could you share the rationale (e.g., compatibility, client behavior expectations, etc.)?
If it is not intended, would you be OK with me submitting a PR to make `/get_backup` return an error when `backup_name` is not found?

Thanks!

huanghaoyuanhhy:
Part of #1171
Fixes #890
Related to #742

## What

`app` gains two usecases where the server handler used to run one fused read:

- `GetBackup` reads the artifact half: the persisted meta in backup storage. `Execute(ctx, name)` takes the bare backup name — a lone selector needs no request struct, same as `GetRestore` — and returns the meta and the meta dir size; a name with nothing behind it is the new `ErrBackupNotFound` sentinel.
- `GetBackupTask` reads the job half: the task manager's view, mirroring `GetRestore`. `Execute(ctx, GetBackupTaskRequest{Name, ID})` — the ID wins when both are set, because only the ID can pin down one of several jobs that shared the name; an unknown selector is `taskmgr.ErrTaskNotFound`.

The two stay separate because they are two resources. v2's API splits "get backup" and "get backup task" into their own endpoints, and with this split each endpoint maps to exactly one usecase. The v1 `get_backup` contract merges both halves into one response, so the merge moves out of the app layer into the v1 handler:

- the artifact is the source of truth; the job overlays progress when one is known
- an in-flight or failed job that has persisted nothing yet answers with the job view alone
- a job that reports success while its artifact is missing is an error — the backup it claims to have produced is gone
- a selector with neither half behind it is an error, not success with zeros (#890)
- a completed backup queried after a restart returns its full payload (#742)

## Wiring

- usecase inputs carry selectors only; overriding where backups live is a config change, made on the `v2.Config` the usecase is built from. The v1 `path` query parameter follows that rule: the handler forks the loaded config with `backup.storage.rootPath` set to it before constructing `GetBackup`, instead of sending the path in as a call argument. The fork is reload-equivalent (#1215), so unset `backup.storage` leaves still cascade from `milvus.storage`, and the server's own config is never mutated.

- `NewGetBackup(ctx, params)` builds only the storage client; `NewGetBackupTask(taskMgr)` takes the process-local manager as a parameter, keeping `taskmgr.DefaultMgr()` out of the usecase layer. Job state needs no config and its construction cannot fail, same as `GetRestore`.
- the `get` CLI reads the artifact half only: a CLI invocation shares no process with the server, so there is never a job to overlay. It no longer imports `internal/taskmgr`, `internal/meta` or `internal/storage`.

## Tests

- `app/get_backup_test.go`: the artifact branches — meta read, not-found sentinel, path override, exist-check and size-probe failures
- `app/get_backup_task_test.go`: the job branches — by name, by ID, ID-wins-over-name, unknown selectors, empty request
- `core/server/get_backup_test.go`: httptest cases over the v1 merge contract — merged rendering, path parameter forks the config, artifact without job, in-flight job without artifact, success-job-without-artifact failure, neither-half not-found, ID-to-name resolution, parameter rejection before either usecase runs, request_id handling, constructor and read errors mapped to `Fail`

### DIFF
--- a/app/get_backup.go
+++ b/app/get_backup.go
@@ -1,0 +1,75 @@
+package app
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/zilliztech/milvus-backup/core/proto/backuppb"
+	v2 "github.com/zilliztech/milvus-backup/internal/cfg/v2"
+	"github.com/zilliztech/milvus-backup/internal/meta"
+	"github.com/zilliztech/milvus-backup/internal/storage"
+	"github.com/zilliztech/milvus-backup/internal/storage/mpath"
+)
+
+// ErrBackupNotFound reports a selector with no artifact behind it. Callers
+// that merge the artifact with the job half (the v1 get_backup contract) use
+// it to tell "no backup persisted yet" apart from a storage failure.
+var ErrBackupNotFound = errors.New("backup not found")
+
+// GetBackup reads one backup artifact: the persisted meta in the backup
+// storage. The job that produced it is a different resource with its own
+// usecase, GetBackupTask — the two are read separately so transports with a
+// split API (v2) map one endpoint to one usecase, and a transport with a
+// merged contract (v1) does the merging itself.
+type GetBackup struct {
+	cli      storage.Client
+	rootPath string
+}
+
+// NewGetBackup builds the usecase from config, creating the backup storage
+// client itself so the transports never import internal/storage. The client
+// is created per call; sharing one across calls is a lifecycle decision this
+// layer deliberately does not make.
+func NewGetBackup(ctx context.Context, params *v2.Config) (*GetBackup, error) {
+	cli, err := storage.NewBackupStorage(ctx, params)
+	if err != nil {
+		return nil, fmt.Errorf("app: %w", err)
+	}
+
+	return &GetBackup{
+		cli:      cli,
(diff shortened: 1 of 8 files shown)


lentitude2tk (approved):
/LGTM

PR #1204, opened 2026-09-05
app: extract the get restore usecase

huanghaoyuanhhy:
Part of #1171.

## What

`core/app` gains the `GetRestore` usecase: one struct holding only the task manager, with `Execute(ctx, id)` as the transport-neutral action — the shape #1171 sketches for it ("takes only the task manager and never touches storage"), and the thinnest usecase in the layer so far. Construction cannot fail, so unlike its siblings `NewGetRestore` takes no config and returns no error.

The `/api/v1/get_restore` handler narrows to request conversion, the usecase call, v1 rendering through `pbconv.RestoreTaskViewToResp`, and error-to-code mapping, through a `newGetRestore` constructor hook on the server config that tests stub out — the same seam list got in #1182 and delete in #1200.

- `Execute` returns the task manager's `RestoreTaskView` directly, with no app-defined wrapper: a restore has no artifact half to merge (the backup a job produces is read through GetBackup), so unlike #1203's `BackupView` a view struct would only rename the type
- `NewGetRestore` takes the task manager as a parameter instead of reading `taskmgr.DefaultMgr()` itself; the server wiring passes the process-local manager, keeping the global reference out of the usecase layer
- overlap with the open #1203 is kept minimal on purpose: `cmd/` is untouched (get_restore has no CLI path — the get_backup CLI is #1203's file), so the only shared file is `core/server/option.go`, where both PRs each add one field plus one default, a mechanical conflict

## Semantics unchanged

Request handling keeps the historical contract exactly: request_id defaults to the header value or a fresh UUID, an empty id answers `Fail` with "empty restore id" (deliberately not `Parameter_Error`), an unknown task ID is an error mapped to `Fail` rather than a silent success, and success renders the same `RestoreBackupResponse` payload as before.

Responses now go through `writeResponse` like the sibling handlers; the wire shape and status are identical, the difference is that failures are now also logged server-side.

One pre-existing oddity recorded here rather than fixed: the old handler's two error branches had their messages swapped — a non-not-found error answered the hardcoded "restore id not exist in task manager", while a not-found error answered with its own text. `Mgr.GetRestoreTask` can only return `ErrTaskNotFound` or nil, so the "other error" branch was unreachable dead code and folds away; the reachable not-found path still answers `Fail` with the error text, now carrying the `app: get restore task <id>:` prefix the layer's error convention adds.

## Tests

- `app/get_restore_test.go`: a known ID returns the task view; an unknown ID (the post-restart case) errors wrapping `taskmgr.ErrTaskNotFound`
- `core/server/get_restore_test.go`: httptest cases over the handler contract — renders the view through the v1 shape, empty id rejected with `Fail` before the usecase runs, request_id generated and forwarded, constructor and execute errors mapped to `Fail`

### DIFF
--- a/app/get_restore.go
+++ b/app/get_restore.go
@@ -1,0 +1,42 @@
+package app
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/zilliztech/milvus-backup/internal/taskmgr"
+)
+
+// GetRestore reads the state of one restore job. Restore jobs keep all their
+// state in the task manager, so this is the thinnest usecase in the layer:
+// unlike its sister GetBackup it never touches storage, which also means
+// there is nothing to construct and construction cannot fail.
+type GetRestore struct {
+	taskMgr *taskmgr.Mgr
+}
+
+// NewGetRestore builds the usecase on the process-local task manager. The
+// manager's state dies with the process — the existing get_restore contract:
+// after a restart every task ID answers not-found.
+func NewGetRestore() *GetRestore {
+	return &GetRestore{taskMgr: taskmgr.DefaultMgr()}
+}
+
+// RestoreView is the job half alone. A restore has no artifact half of its
+// own — the backup a job produces is read through GetBackup — but it still
+// travels as an app-defined struct so transports never name the task
+// manager's types.
+type RestoreView struct {
+	Task taskmgr.RestoreTaskView
+}
+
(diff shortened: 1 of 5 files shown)


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

Issue #1124, opened 2026-07-27
[Enhancement]: Introduce an HTTP API v2 separating backups from jobs


huanghaoyuanhhy:
## What would you like to be added?

A second HTTP API, mounted at `/api/v2`, speaking the same dialect as Milvus's RESTful v2, that fixes five problems the v1 API cannot fix without breaking its own contract. v1 stays mounted and frozen for one release, then is removed together with its proto messages, the deprecated `github.com/golang/protobuf` dependency, and the `sed` post-processing in the code generation scripts.

This is the API-side counterpart of the configuration work in #1089. It closes #935 and #890, implements the split proposed in #816, and is the surface resumable restore (#1071, currently on `breakpoint-beta`) should land on rather than the one it had to work around.

| Problem | Where it shows |
| --- | --- |
| [Backup state and job progress are one object](#1-backup-state-and-job-progress-are-one-object) | #816, #742, #890, #906 |
| [Four naming conventions in one API](#2-four-naming-conventions-in-one-api) | #935 |
| [A new request field per configuration key](#3-a-new-request-field-per-configuration-key) | #877, #881, #1071 |
| [The proto pays for nothing](#4-the-proto-pays-for-nothing) | #1071 |
| [A long-running job has no lifecycle](#5-a-long-running-job-has-no-lifecycle) | #1071, #941 |

#### 1. Backup state and job progress are one object

`BackupInfoBrief` is a single message whose own field comments read `// from task` and `// from backup meta`. The two halves have different lifetimes: a backup is a directory in object storage that outlives the process, a task is an entry in an in-memory map that does not survive a restart. Merging them produces failures that look unrelated but are all the same defect.

A backup name that does not exist returns success with an empty payload (#890):

```console
$ curl -XGET "http://localhost:8080/api/v1/get_backup?backup_name=nonexist"
{"msg":"success","data":{"state_code":0,"progress":0,"size":0,"meta_size":0}}
```

The cause is in `getByName`: storage is read only when an in-memory task exists **and** that task succeeded. No task, no read, empty brief, `success`.

The same branch hides backups that do exist. After a server restart the task map is empty, so `get_backup` for a backup sitting in the bucket returns the same empty payload, and `state_code: 0` decodes as `BACKUP_INITIAL` — a completed backup reports itself as pending. That is #742, and it is not a second bug.

[...]

PR #1205, opened 2026-09-05
app: extract the create backup usecase

huanghaoyuanhhy:
Part of #1171

## What

`app` gains the `CreateBackup` usecase: one struct holding the config, the two storage clients, the task manager and the backup root path, constructor-built from config. The action is *create a backup job*: a job is what a create call makes, and the backup artifact is what a successful job leaves behind — the two-resource split of the v2 API (#1124), where `jobs/backup/create` returns the job and `backups/describe` reads the artifact.

- `Start(req)` is the synchronous act of starting a job — it assembles `backup.TaskArgs` and calls `backup.NewTask`, which registers the job in the task manager, and returns the registered job ready to run behind a small `BackupJob` interface
- `Execute(ctx, req)` runs the job synchronously and answers with the task manager's view of it: id, state, progress and the rest of the job half — the shape v2's `jobs/backup/create` responds with. It never reads the persisted meta; a transport whose contract merges the two resources (v1) assembles what it needs itself

Both transports call it:

- the `create` CLI command (`cmd/create`) builds `app.NewCreateBackup` + `Execute`; flag parsing, name defaulting, validation and the printed output stay in `cmd`
- the `/api/v1/create` handler (`core/server`) renders the pb request into `app.CreateBackupRequest` — keeping the deprecated pb fields (`db_collections`, `collection_names`, `force`, `meta_only`), name validation and the error-to-code mapping — through a `newCreateBackup` constructor hook on the server config that tests stub out, the same seam list and delete got in #1182 and #1200
- the v1 `async` flag stays a server concern: the handler calls `Start` and spawns the goroutine itself (unchanged logging on failure), while taskmgr registration stays where it was, inside `backup.NewTask`. `core/backup` is not touched
- `NewCreateBackup` takes the task manager as a parameter instead of reading `taskmgr.DefaultMgr()` itself; `cmd/create` and the server wiring pass the process-local manager, keeping the global reference out of the usecase layer

Rebased on top of the config fork (#1215): the v1 `backup_root_path` field is applied as a config override, not a request field. The handler forks the loaded config with `backup.storage.rootPath` as one more override layer — the same move `get_backup` makes for its `path` parameter in #1203 — and the forked params are what `NewCreateBackup` builds the usecase from, so `CreateBackupRequest` carries no root path and the usecase reads it from its params. The fork is reload-equivalent, so unset `backup.storage` leaves still cascade from `milvus.storage`, and the server's own config is never mutated. The artifact directory for a given request is the same as before the rebase.

## Semantics unchanged

The v1 wire shape is kept, including its quirks:

- the sync success response still carries only `code` and `msg` — no payload and no `request_id`. The usecase now answers with the job view alone, and the v1 contract needs nothing from it, so the handler discards it; anything a v1 response does need, the handler assembles itself
- `request_id` defaulting, error-to-code mapping (`Parameter_Error` for a bad name, `Fail` for everything else, `request_id` echoed on failures), filter/strategy/format precedence and the deprecated-field fallbacks are unchanged
- async behavior is unchanged: same response message (`create backup is executing asynchronously`), same `request_id` echo, task still registered synchronously in the default task manager before the goroutine starts

Two pre-existing bugs were found and are **not** fixed here:

- empty `backup_name` still panics in `backup.ValidateName` and lands on gin's recovery as a 500, exactly as before

Two deliberate behavioral deltas, both limited to error paths:

- the sync path used to fall through from a `backup.NewTask` error into `task.Execute` on a nil task — a guaranteed panic. The usecase returns that error instead, so a job whose registration fails now answers `Fail` with the error message
- the old sync path read the task view, the persisted meta and the meta size after a successful run and answered `Fail` when any of those reads failed — even though the backup had already succeeded — all to build a brief it then discarded. That read only ever fed the discarded payload, so it is gone together with the payload: a successful backup now always answers success

## Tests

- `app/create_test.go`: `Start` registers the job in the task manager and refuses a second live job for the same backup name; `Execute` propagates the start failure; the artifact dir is the configured root path plus the option's name
- `core/server/create_test.go`: httptest cases over the handler contract — sync runs through the usecase and the success response carries no payload (quirk pinned), request_id generated and forwarded, invalid name rejected with `Parameter_Error` before the usecase runs, constructor/build/execute/start errors mapped to `Fail`, async starts the job and returns immediately, `backup_root_path` forks the loaded config while the server's own config stays on the default root path (and without the field the config travels to the usecase untouched); plus the pb→option rendering incl. deprecated `force`/`meta_only` and the three filter sources

### DIFF
--- a/app/create.go
+++ b/app/create.go
@@ -1,0 +1,170 @@
+package app
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/zilliztech/milvus-backup/core/backup"
+	"github.com/zilliztech/milvus-backup/core/proto/backuppb"
+	v2 "github.com/zilliztech/milvus-backup/internal/cfg/v2"
+	"github.com/zilliztech/milvus-backup/internal/meta"
+	"github.com/zilliztech/milvus-backup/internal/storage"
+	"github.com/zilliztech/milvus-backup/internal/storage/mpath"
+	"github.com/zilliztech/milvus-backup/internal/taskmgr"
+)
+
+// BackupJob is one registered create-backup run, ready to execute. Starting a
+// job and running it are separate steps so a transport that executes jobs
+// asynchronously can decide itself where the goroutine goes.
+type BackupJob interface {
+	// Run executes the job. The outcome is also recorded in the task manager,
+	// where get_backup and the task API read it from.
+	Run(ctx context.Context) error
+}
+
+// CreateBackup starts one backup job that writes an artifact into the backup
+// storage, copying from the milvus storage.
+type CreateBackup struct {
+	params *v2.Config
+
+	milvusStorage storage.Client
+	backupStorage storage.Client
+
+	taskMgr  *taskmgr.Mgr
+	rootPath string
+}
+
+// NewCreateBackup builds the usecase from config, creating both storage
+// clients itself so the transports never import internal/storage. The clients
+// are created per call; sharing them across calls is a lifecycle decision
+// this layer deliberately does not make.
+func NewCreateBackup(ctx context.Context, params *v2.Config) (*CreateBackup, error) {
+	backupStorage, err := storage.NewBackupStorage(ctx, params)
+	if err != nil {
+		return nil, fmt.Errorf("app: %w", err)
+	}
+
+	milvusStorage, err := storage.NewMilvusStorage(ctx, params)
+	if err != nil {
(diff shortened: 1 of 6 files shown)


lentitude2tk (approved):
/LGTM

lentitude2tk (approved):
/LGTM

PR #1206, opened 2026-09-05
app: extract the check usecase

huanghaoyuanhhy:
Part of #1171.

## What

The `app` package gains the `Check` usecase: one struct holding the Milvus gRPC client and the two storage clients, constructor-built from config, with `Execute(ctx, output io.Writer)` as the transport-neutral action. Both transports call it and `core/check` goes away:

- the `check` CLI command (`cmd/check`) calls `app.NewCheck` + `Execute` instead of assembling a gRPC client, two storage clients, and a `core/check.Task` itself; printing the effective configuration and passing `cmd.OutOrStdout()` stay in the command
- the `/api/v1/check` handler (`core/server`) does the same, through a `newCheck` constructor hook on the server config that tests stub out — the same seam list got in #1182 and delete in #1200
- `core/check` had no callers left, so the package folds into the usecase. Unlike `core/del` it was never a shared implementation: `git grep` shows exactly two importers, both transports this PR converts

The usecase stores the three config values it needs (Milvus root path, backup root path, transfer mode) instead of the whole config, and takes the report writer as an `Execute` parameter — the report is plain text by design; how a transport presents it (stdout, HTTP body) is the transport's call.

## Semantics unchanged

The five checks still run in the same order — Milvus connect, Milvus storage, backup storage, write-and-copy round trip, report — and still stop at the first failure, with the same deferred cleanup of the two check objects. The report text (version line, empty-root warning, success line) is byte-identical, and both transports still lead their output with the effective configuration before any client is constructed. Error-to-response mapping stays in the handler (500 `{"error": ...}`), the CLI's exit semantics stay `cobra.CheckErr`. The only textual change is the error prefix moving from `check:` to `app:` inside the wrapped chains, as accepted for this layer.

## Tests

- `app/check_test.go`: mocked gRPC + storage clients over the usecase contract — reports version and success; warns when the Milvus root is empty; fails at each step boundary (Milvus unreachable, Milvus root unreadable, backup root unreadable, write failure, copy listing failure) with nothing past that step running; the written check object is still cleaned up when the copy fails
- `core/server/check_test.go`: httptest cases over the handler contract — 200 body renders config first then the usecase report, constructor and execute errors map to 500 `{"error": ...}`


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

PR #1207, opened 2026-09-05
app: extract the restore usecase

huanghaoyuanhhy:
/kind improvement

Part of #1171.

## What

`app` gains the restore family as two usecases, `Restore` and `RestoreSecondary`, shaped like their siblings — create (#1205) and get backup (#1203). Both transports call them; only the per-transport input grammar and the v1 rendering stay behind:

- `app.NewRestore(ctx, params, taskMgr)` creates both storage clients itself, so transports never import `internal/storage`, and resolves the root path from config. `NewBackupStorage` also creates the backup bucket when it is missing — a restore may target a bucket nothing has written yet, which the old code arranged per request and the constructor now arranges per construction. `app.NewRestoreSecondary` is the same for a secondary restore. The clients are created per call, as the other usecases' are.
- `app.Restore.Start(ctx, req)` validates the request against the backup storage — the backup must exist and its meta must be readable, because a not-found backup is the caller's mistake and has to be answered before the job is registered — then builds the `restore.Task`; task creation is what registers the job, exactly as before. It returns an `app.RestoreJob`, a one-method `Run(ctx) error` interface like create's `BackupJob`, so the transports drive the job without importing the task layer. The async split follows the issue's rule: the action is synchronous, the goroutine is the HTTP server's deployment concern. A not-found backup is the shared `app.ErrBackupNotFound` sentinel from get_backup, wrapped with the name the same way.
- The v1 `bucket_name` and `path` fields are config overrides, not request options: the handler forks the loaded config with them (`backup.storage.bucketName` / `backup.storage.rootPath`) and builds the usecase from the fork, exactly as get_backup does for its path and create for its root. The override fields therefore do not exist on the usecase request, and there is no per-Start client creation — the seams are the standard `func(ctx, params) (uc, error)` shape.
- `core/server` gets `newRestoreBackup` / `newRestoreSecondary` constructor hooks on the server config (the same seam list create and the rest got), defaulting to `app.NewRestore` / `app.NewRestoreSecondary` with the process-local task manager; handler tests stub them.
- The response's task view is read through the get-restore usecase — the same usecase `/get_restore` answers from — via a small `restoreTaskView` helper. There is no `TaskView` method on the restore usecases and no view alias: the job half has one reader.
- The `/restore` handler narrows to: bind the pb request, default request id and task id, validate, translate the pb grammar into `app.RestoreRequest` (plan and option are parse products of pb-only fields, including the deprecated `db_collections`, so they are built here), fork the config when the request overrides bucket or path, build the usecase, start it, then run the job synchronously or launch it in its own goroutine with the same log line as before, and render. The same narrowing applies to `cmd/restore` and `cmd/restore/secondary`, which keep their flags, their plan building and the duration print.
- `core/restore` (and its `secondary` subpackage) is untouched; the app layer injects and drives it. `RestoreRequest.Plan`/`.Option` deliberately use the task layer's own vocabulary: each transport translates its grammar into them (flags vs pb fields), and the two grammars do not map onto each other, so there is nothing transport-neutral to put there instead.

## Semantics unchanged

Sync and async execution, the `restore_<timestamp>` task id default, `Parameter_Error` vs `Fail` mapping, the per-endpoint response quirks (the v1 `/restore` failures before the task exists respond without a request id; `/restore_secondary` failures respond with it), the async success message, the goroutine log lines, and the secondary defenses (full-meta read, index-extra preconditions inside `secondary.NewTask`) are all preserved. The `use bucket name from request` / `use path from request` log lines move with the overrides into the handler's fork block.

Five declared deltas, all smaller than the noise they replace:

1. Error precedence on `/restore` when both the request grammar and storage are broken: the old handler built the plan after the storage clients came up, so storage errors won; the plan is now translated before the usecase is built, so a malformed plan (bad `collection_renames`, unparsable `db_collections`) is reported first. The CLI already had the plan-first order, so the two transports now agree. Error codes are unaffected (both are `Fail`).
2. The two `/restore_secondary` error paths that previously responded without a request id (storage failure during the existence check, and backup not found) now respond with it, because every `Start` failure flows through one response; construction failures do too. Every other `/restore_secondary` failure already carried the id; this removes the inconsistency rather than adding one.
3. On `/restore_secondary` the target cluster's storage client is now built before the backup's existence is checked, so a broken target storage wins over a missing backup where the old order reported the missing backup. `/restore` is unaffected — both clients already came up before the existence check there.
4. The not-found answer is now the `ErrBackupNotFound` sentinel, so its text is `app: backup <name>: backup not found` (was `backup <name> not found`), matching get_backup.
5. `cmd/restore` error strings change with the moved assembly (`create backup storage: ...` → `app: create backup storage client: ...`); the `NewBackupStorage`-vs-`NewClient`+`CreateBucketIfNotExist` difference between the CLI and the old server handler collapses into one path that creates the bucket either way.

Also noted, not changed: the v1 `/restore` and `/restore_secondary` endpoints ignore the `request_id` header that their swagger comments advertise; the id comes from the body field only. Fixing that would be a behavior change, so it is left as is.

## Tests

- `app/restore_test.go`: mock-client coverage for both usecases over the struct-literal construction (the constructor creates real clients, so tests skip it as create's do) — `Start` assembles and registers the job (the view becomes visible in the usecase's own task manager with the initial state), backup not found (`ErrBackupNotFound`), existence-check failure, meta-read failure, task refusing an unknown format
- `core/server/restore_test.go`: the plan/option/filter translation tests carry over unchanged (the functions moved nowhere); new httptest handler-contract cases for both endpoints over the stubbed usecases — success and async paths, invalid body, missing name/cluster ids rejected before the usecase runs, request id and task id defaulting and forwarding, bucket and path forking the config (and no fork without them), not-found mapped to `Parameter_Error`, construction/start/run/get-restore errors mapped to `Fail`
- `internal/storage/factory_test.go`: mapping tests for the shared `storageConfig` and its two wrappers — the section a usecase's (possibly forked) params carry is what the client actually runs on, so this pins the field mapping and that `BackupStorageConfig` reads the backup section rather than the milvus one, closing the one link the handler and usecase tests deliberately stub around

### DIFF
--- a/app/restore.go
+++ b/app/restore.go
@@ -1,0 +1,186 @@
+package app
+
+import (
+	"context"
+	"fmt"
+
+	"go.uber.org/zap"
+
+	"github.com/zilliztech/milvus-backup/core/restore"
+	"github.com/zilliztech/milvus-backup/core/tasklet"
+	v2 "github.com/zilliztech/milvus-backup/internal/cfg/v2"
+	"github.com/zilliztech/milvus-backup/internal/log"
+	"github.com/zilliztech/milvus-backup/internal/meta"
+	"github.com/zilliztech/milvus-backup/internal/storage"
+	"github.com/zilliztech/milvus-backup/internal/storage/mpath"
+	"github.com/zilliztech/milvus-backup/internal/taskmgr"
+)
+
+// BackupNotFoundError reports a restore request naming a backup that has no
+// artifact in the backup storage. Transports map it to their parameter-error
+// code: the caller asked for something that is not there.
+type BackupNotFoundError struct {
+	Name string
+}
+
+func (e *BackupNotFoundError) Error() string {
+	return fmt.Sprintf("app: backup %s not found", e.Name)
+}
+
+// RestoreTaskView aliases the task manager's view of one restore job. It is
+// what a transport renders into its own wire shape; the alias exists so the
+// transports need not import internal/taskmgr just to name the type.
+type RestoreTaskView = taskmgr.RestoreTaskView
+
+// RestoreJob is one assembled restore job: its task exists and the task
+// manager knows it, but nothing has run yet. The transport decides how the
+// job executes — synchronously via Execute, or in the transport's own
+// goroutine when it restores asynchronously.
+type RestoreJob interface {
+	// Execute runs the job to completion.
+	Execute(ctx context.Context) error
+
+	// TaskID identifies the job in the task manager; TaskView takes it.
+	TaskID() string
+}
+
+// Restore restores a backup into the target Milvus.
+type Restore struct {
+	params *v2.Config
+
+	// newBackupStorage and newMilvusStorage build the two clients one restore
+	// runs on. They are per-Start calls, not constructor state: the backup
+	// client depends on the request's bucket override, which no constructor
+	// knows. They are fields so tests inject mocks instead of a real backend.
+	newBackupStorage func(ctx context.Context, bucketName string) (storage.Client, error)
+	newMilvusStorage func(ctx context.Context) (storage.Client, error)
+}
+
+// NewRestore builds the usecase from config.
+func NewRestore(params *v2.Config) *Restore {
+	return &Restore{
+		params: params,
+		newBackupStorage: func(ctx context.Context, bucketName string) (storage.Client, error) {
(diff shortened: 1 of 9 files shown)


lentitude2tk (approved):
/LGTM

PR #1210, opened 2026-09-08
cfg: add Config.Fork for map-based overrides on a frozen copy

### LINKED ISSUE #1209: cfg: derive a resolved v2 config with map-based overrides (Fork)
## Motivation

The whole program shares a single `*v2.Config` pointer, loaded once via `loader.Load` and treated as read-only afterwards. There is no way to derive a modified copy: per-request tweaks (e.g. the restore handlers rewriting `bucketName`/`path` into a derived `storage.Config`, `core/server/restore.go:137-153`) bypass the config tree entirely, so each call site re-invents its own ad-hoc override and the config tree no longer describes what is actually running.

The first consumer is the GetBackup usecase (#1203), which needs to honor a per-request `path` parameter without mutating the shared config.

## Proposal

A fork is **reload-equivalent**: the same load, plus an extra layer of overrides. Not a frozen leaf copy.

- `v2.Config` retains the `param.Source` it was loaded from (unexported pointer field; `param.Walk` skips pointer fields, so display/render/defaults are unaffected).
- `param.Source` gains `WithOverrides(map[string]string) *Source`: clone the source and merge the override map (later wins; the original source is never mutated).
- `Fork(overrides)` = `checkForkKeys` (unknown keys are **errors**, with the v1 migration hint from `legacy.go`) → `Resolve(src.WithOverrides(m))` → `Validate()` → attach the merged source. Fork-of-fork accumulates naturally.
- Override map semantics = `--set`: dotted lower-case config keys or credential env names, string values via the same `parseValue` path, lists comma-separated.

Semantics that follow from reload-equivalence, accepted deliberately:

- Re-resolution re-reads the process environment (static in practice) and re-runs `backup.storage` inheritance — overriding `milvus.storage.*` cascades into unset `backup.storage` leaves, exactly as if the process had been restarted with an extra `--set`.
- Provenance is reproduced by re-resolution, not copied: env-sourced leaves keep their env stamp, file leaves their file stamp, first-load overrides their override stamp.
- `Fork(nil)` is a validated independent copy; the receiver is never mutated.
- Fork on a config that was never loaded from a source is an explicit error.

## Dependency

v1-loaded configs currently have no source. #1211 moves v1 translation to the source level; once it lands, every config carries a source and Fork works uniformly. Until then Fork errors on v1-loaded configs.

## Design history

An earlier iteration implemented Fork as a frozen copy (render receiver to in-memory YAML, re-resolve, then reflectively copy back untouched leaves to restore values/provenance and suppress inheritance). Superseded: reload-equivalence is simpler, matches operator intuition about `--set`, and needs no copy-back machinery. See #1210 (closed unmerged) for that iteration.

huanghaoyuanhhy:
Fixes #1209

Related: #1203

## What

Adds `func (c *Config) Fork(overrides map[string]string) (*Config, error)` to `internal/cfg/v2`: derive an independent copy of a resolved configuration with a `--set`-style override map applied. The first consumer is the GetBackup usecase (#1203), which needs to honor a per-request `path` parameter without mutating the shared config.

## Semantics

- **Frozen copy.** Every leaf keeps the receiver's resolved value unless the override map names it. The process environment is not consulted on the fork pass, and `backup.storage` does **not** re-inherit from an overridden `milvus.storage` — overrides apply to leaves, nothing cascades.
- **Provenance preserved.** Untouched leaves keep their original `Used` stamps (including `override`/`env` stamps from the parent load); overridden leaves get a fresh `SourceOverride` stamp.
- **`--set` semantics for the map.** Dotted lower-case config keys or credential environment names, string values parsed by the same `parseValue` path, lists comma-separated.
- **Fail-fast.** Unknown keys are errors (with a v1 migration hint), unparseable values error, and the derived configuration is re-`Validate()`d before returning. `Fork(nil)` yields a validated deep copy; the receiver is never mutated.

## How

Render the receiver to in-memory YAML → resolve a fresh tree from a new `param.NewSourceYAML` (in-memory, environment disabled) → walk both trees in parallel and copy back every leaf the map does not name, restoring value and provenance → `Validate()`. Re-resolution is only the application mechanism for the overridden leaves; the copy-back is what makes the fork differ from its parent at exactly the overridden keys.

## Tests

`fork_test.go` covers the frozen-copy equality, single-leaf override (by config key and by env name), the no-cascade guarantee, unknown/v1/removed keys, unparseable values, validation failure, post-load environment isolation, provenance preservation, list overrides, parent immutability, and deep-copy independence.

### DIFF
--- a/internal/cfg/param/source.go
+++ b/internal/cfg/param/source.go
@@ -23,20 +23,29 @@ type Source struct {
 	override     map[string]string
 	overrideKeys []string
 
+	// noEnv disables environment lookups. A source built from an
+	// already-resolved rendering (a config fork) must not pick up process
+	// variables on the second pass: its YAML layer is the resolved state.
+	noEnv bool
+
 	configFile map[string]any
 }
 
-// NewSource reads configPath and flattens it into dotted lower-case keys.
-// An empty configPath yields a source backed by overrides and env only.
-func NewSource(configPath string, overrides map[string]string) (*Source, error) {
+func newSource(overrides map[string]string) *Source {
 	s := &Source{override: make(map[string]string, len(overrides)), configFile: map[string]any{}}
-
 	for k, v := range overrides {
 		s.override[strings.ToLower(k)] = v
 		s.overrideKeys = append(s.overrideKeys, k)
 	}
 	slices.Sort(s.overrideKeys)
 
+	return s
+}
+
+// NewSource reads configPath and flattens it into dotted lower-case keys.
+// An empty configPath yields a source backed by overrides and env only.
+func NewSource(configPath string, overrides map[string]string) (*Source, error) {
+	s := newSource(overrides)
 	if configPath == "" {
 		return s, nil
 	}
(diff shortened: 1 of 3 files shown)


huanghaoyuanhhy:
Closing unmerged: the design was revised on #1209 after review discussion. The frozen-copy + copy-back implementation here is superseded by reload-equivalent semantics (retain the source, re-resolve with merged overrides), which also depends on #1211 landing first so v1-loaded configs carry a source. Will resubmit on top of that work.

PR #1212, opened 2026-09-08
migrate: translate v1 configs at the source level

### LINKED ISSUE #1211: migrate: translate v1 configs at the source level instead of field-by-field
## Motivation

The v1→v2 compatibility path today resolves the v1 schema and then copies it into a v2 config field by field (`migrate.translate`, ~250 lines). That function is a maintenance hazard: every new v2 field must be mirrored there by hand, and forgetting silently leaves v1 users with zero values.

It also means a v1-loaded configuration is the one config in the program with no underlying `param.Source` — which blocks the config Fork proposal (#1209), where every config carries the source it was loaded from so a fork is a re-resolution with extra overrides.

## Proposal

Move the version boundary from "resolved config" to "source": translate the flattened v1 key map into a flattened v2 key map, then let `v2.LoadFrom` do all resolution. Both schema versions then flow through the same Source→Resolve→Validate pipeline, and the v1 schema package becomes dead code.

Concretely:

- A pure rename table for the ~50 mechanical keys, plus a small transform registry for the ~7 semantic ones:
  - `milvus.tlsMode`: 0/1/2 → disabled/server/mutual
  - `milvus.etcd.endpoints`: comma-separated string → list
  - `minio.useIAM: true` → `auth.type: iam`
  - `minio.gcpCredentialJSON` → `auth.credentialsFile` **plus** `auth.type: serviceAccount` (one-to-two)
  - `minio.accessKeyID` → `accountName` when the provider is azure, else `auth.accessKeyID` (cross-key: needs the translated `storagetype` first)
  - provider spelling normalization (ali/alicloud → aliyun, ...)
  - `storage.storageType` / `minio.storageType` / `minio.cloudProvider` map to one field; honor v1 declaration-order precedence, not map iteration
- **Env aliasing in `param.Source`**: v1 environment names (`MINIO_SECRET_KEY`, ...) must keep resolving — the translated source maps each v2 credential env name to its v1 spelling, tries both, and stamps the one actually hit. Without this, v1 users' env credentials silently stop working.
- The override map (`--set`) is translated through the same table, so v1-spelled overrides keep their `override` provenance stamp.
- **Defaults parity audit**: keys unset in v1 fall to v2 defaults after translation. Most defaults already agree (`minioadmin`, `a-bucket`, `files`, ...); where v1 semantics differ — notably the backup-side rootPath inheritance wording — inject the v1-resolved default into the translated source so behavior does not drift.
- The `config migrate` CLI report is rebuilt on the same table: the translation walk itself yields the renamed/transformed/dropped outcomes that are the report's data source.
- Delete the v1 schema package (`internal/cfg/cfg.go`, `load.go`, `storage_types.go`) once nothing resolves it anymore.

## Outcome

- `migrate.translate` field-by-field copy is deleted; adding a v2 field no longer has a hidden second site.
- Every config in the program carries its source, so Fork (#1209) works uniformly for v1- and v2-loaded configs with no special case.
- Provenance survives translation at full fidelity: file → file, override → override, env → env.

huanghaoyuanhhy:
## What

Moves the v1→v2 version boundary to the source, as proposed in the issue. `param.Source` is now the single materialization layer: it reads the config file, snapshots the environment, and takes the `--set` overrides once, and nothing is read after construction. A v1 file is translated at the key level — the flattened v1 key map is renamed into the v2 key space — and both schema versions resolve through the same Source → Resolve → Validate pipeline. v2 is the only config interface, and every loaded config carries the `param.Source` the config fork proposal (#1209) needs.

The ~250-line field-by-field `migrate.translate` copy is gone, so adding a v2 field no longer has a hidden second site that silently leaves v1 users with zero values.

## How

- **`param.Source`** materializes three layers: `--set` overrides, an environment snapshot, and the flattened config file — whose values are `Input` structs carrying the value, the source kind, and the key or variable name the value was spelled with. Resolution stamps provenance straight from the entry, so param stays schema-agnostic; the new `v1 env` / `v1 config` source kinds only name what a translation produced.
- **`migrate/keymap.go`** — the machine-readable v1 field table: one entry per logical v1 parameter with its config keys and env names in v1 declaration order (first hit wins). `v2/legacy.go` keeps the prose table for "you used a v1 name in a v2 file" warnings.
- **`migrate/translate.go`** — the translator. It gathers each field's winning hit across the v1 precedence layers (override > env > file) and emits one entry per field into the v2 key space, stamped with the v1 spelling and the source kind it came from. Settling precedence per field is what lets the semantic mappings see exactly what v1 resolution saw: tlsMode 0/1/2 → enum (with the mTLS downgrade settled), etcd endpoints string → list, useIAM → `auth.type: iam`, gcpCredentialJSON → credentialsFile + `auth.type: serviceAccount`, provider-conditional accessKeyID/secretAccessKey targets (azure account name/key vs static pair), provider spelling aliases, and the three storagetype keys with declaration-order precedence. The translator vouches `configVersion: v2` for its output, so `v2.LoadFrom` runs unchanged.
- **Bolted on, not woven in**: `internal/cfg/migrate` is the only package that knows the v1 key space, and the loader's v1 branch is the only caller on the load path. Removing v1 support means removing that package, the branch, and the two v1 source kinds — nothing in param, v2, or the rest of the tree mentions v1.
- **Backup root path**: v1 let an unset backup root path inherit the Milvus root path *unconditionally* (the v1 default `files` is non-empty, so the inheritance always fired), while v2 defaults to `backup`. The translator injects the inherited value explicitly, so backups keep landing where v1 put them; the report warns when the Milvus root path was explicitly set.
- **`config migrate`** rebuilds its report from the same walk: renamed / transformed / dropped keys, env deferrals, comments, and env renames. Env-sourced secrets are still withheld from the written file and deferred to the v2 variable, including the backup side when it inherits an env-sourced primary secret. `v2.Resolve` (now exported) resolves without validating so the report can collect the problems `--strict` promotes.
- The v1 schema package (`internal/cfg/cfg.go`, `load.go`, `storage_types.go`) and `param.Defaults`, which only the struct translate used, are deleted.

## Verification

- **Field-table audit**: the machine-readable v1 table was compared against the retired v1 schema mechanically — all 59 parameters match on key spellings, declaration order, and env names; all 43 plain-rename defaults are equal (the four backup-side declared defaults differ only because v1 overwrote them with inherited values at resolve time, which v2's `inherit()` reproduces — rootPath excepted, handled explicitly).
- **41 golden cases** pin the loader: v1 configurations covering providers and aliases, azure/gcpnative/IAM auth, TLS modes, etcd endpoints from every layer, backup inheritance, crossStorage, env and override interactions, were run through the *pre-translation* loader (including the `cmp.Or` corner where an explicitly empty Milvus root path lands the backup side on `backup`) and the goldens committed; the source-level path must reproduce them value for value (`internal/cfg/loader/testdata/golden`, regenerate with `GOLDEN_WRITE=1` against a pre-translation tree).
- The rendered file and the stderr report of `config migrate` are **byte-identical** to the pre-translation binary on a config that trips the comment/warning paths (http.enabled drop, mTLS downgrade, env-secret deferral, inherited root path, crossStorage divergence).
- Full test suite, `-race`, and golangci-lint are clean.

## Intended behavior changes (all surfaced, none silent)

- **Provenance is now truthful** on the v1 path: the struct translate copied values onto a defaulted struct, so `config show` reported everything as `default`; now a value reports the v1 file key / env variable / `--set` spelling it came from, under the `v1 config` / `v1 env` kinds.
- **Unknown v1 file keys are dropped with a warning** instead of silently ignored.
- **A v2-spelled `--set` passes through to a v1 load** instead of being dropped: the program's own override names keep working while a file is still on v1.
- **Azure without any credentials no longer materializes the v1 `minioadmin` default** as the account name/key — the config fails validation instead of carrying a credential that was never real.
- **v2 inheritance actually runs on the v1 path now**: the struct translate bypassed it, so for example a backup side that opts out of IAM did not inherit `auth.endpoint`; through the standard backup-side inheritance it now does (inert under static auth, which never reads it). Same mechanism, bare gcpnative configs inherit the primary credentials file where the old loader path hard-failed validation.

Fixes #1211


lentitude2tk (approved):
/LGTM

PR #1214, opened 2026-09-09
backup: log snapshot lifetime and retry drop in client

huanghaoyuanhhy:
Two changes to the snapshot backup task's drop path:

- **Log the snapshot's lifetime.** `snapshot created` once `CreateSnapshot` succeeds, `snapshot dropped` once the drop succeeds. A snapshot pins its collection's files in Milvus against GC for as long as it exists, so a snapshot that is never dropped is the failure worth seeing; the success path used to be silent, which made a leaked snapshot indistinguishable from a clean backup in the log.
- **Retry the drop in the milvus client.** `GrpcClient.DropSnapshot` now retries while the server refuses the call, and the task calls it once and logs the outcome. The reason for the retry — DataCoord releases the export's pin on its own reconcile tick, not inline with the job's terminal state — belongs with the RPC, and it is how every other retry in the client package already works. The per-attempt warning log goes away with the loop; the final failure is logged once, with the client's error.

Behaviour delta: the wait budget becomes ten attempts with exponential backoff (about 24s worst case) instead of ten attempts one second apart (about 10s). The first retry is still after one second.

/kind improvement

### DIFF
--- a/core/backup/coll_snapshot_task.go
+++ b/core/backup/coll_snapshot_task.go
@@ -17,12 +17,6 @@ import (
 
 const (
 	_snapshotPollInterval = 3 * time.Second
-
-	// A drop right after the export job goes terminal can still be rejected: DataCoord
-	// releases the pin it took for the copy on its own reconcile tick, not inline with
-	// the state change the poll observed.
-	_snapshotDropRetry    = 10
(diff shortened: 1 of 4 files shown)


lentitude2tk (approved):
/LGTM

PR #1215, opened 2026-09-09
cfg: fork a resolved v2 config with map-based overrides

### LINKED ISSUE #1209: cfg: derive a resolved v2 config with map-based overrides (Fork)

huanghaoyuanhhy:
## What

`v2.Config` gains `Fork(overrides)`: derive an independent, validated configuration from a loaded one, with the overrides applied as one more `--set` layer.

Fixes #1209

## How

A fork is reload-equivalent: the same load, plus an extra override layer.

- `v2.Config` retains the `param.Source` it was resolved from (unexported pointer field; `param.Walk` skips pointer fields, so display, render, defaults and key declaration are unaffected). The source is attached in `Resolve`, not `LoadFrom`, so the `config migrate` artifact — which resolves without validating — carries one as well, and the Translate/Migrate parity test keeps comparing whole configurations.
- `param.Source` gains `WithOverrides(map[string]string) *Source`: clone and merge, later wins, and the latest spelling is the one errors quote back. The receiver is never mutated. The env and file layers are never written after construction, so the copy shares them and only the override layer is rebuilt.
- `Fork(m)` = `checkForkKeys` (unknown keys are errors, with the v1 migration hint from `legacy.go`) → `Resolve(src.WithOverrides(m))` → `Validate()` → attach the merged source. Fork-of-fork accumulates naturally, `Fork(nil)` is a validated independent copy, and forking a configuration that was never loaded is an explicit error.

Semantics that follow from reload-equivalence, called out in the issue: overriding a `milvus.storage` leaf cascades into unset `backup.storage` leaves (exactly what an extra `--set` at startup would do), and provenance is reproduced by re-resolution rather than copied — env stays env, file stays file, v1 stays v1.

## Tests

- `param`: the merge itself — later wins, receiver immutable, nil is a plain copy, file and env layers pass through.
- `v2`: a nil copy reproduces values and provenance exactly; an override applies without mutating the receiver; the cascade reaches unset backup leaves; the env snapshot is re-read and a credential may be named by its environment variable; fork-of-fork accumulates; a never-loaded config errors; misspelled, v1 and removed keys error with the migration hint; validation and parse failures propagate.
- `loader`: a v1-loaded configuration forks through its translated source and keeps v1 provenance (`v1 config` stamp with the original v1 key).

The first consumer — the GetBackup usecase honoring its per-request `path` parameter (#1203) — follows as a separate PR once this lands.

### DIFF
--- a/internal/cfg/loader/load_test.go
+++ b/internal/cfg/loader/load_test.go
@@ -7,6 +7,8 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/zilliztech/milvus-backup/internal/cfg/param"
 )
 
 // write puts content in a temp file and returns its path, so a test starts from
@@ -113,6 +115,30 @@ func TestLoad_UnknownVersion(t *testing.T) {
 	require.Error(t, err)
 	assert.ErrorContains(t, err, "v3")
 	assert.ErrorContains(t, err, "not a schema version this build knows")
+}
+
+// A v1 file loads through source-level translation, and the translated source
+// is the one Fork re-resolves — the v1 values ride along, still stamped with
+// the v1 keys they came from.
+func TestLoad_V1FileForks(t *testing.T) {
+	out, err := Load(write(t, `
+milvus:
+  address: milvus-proxy
+minio:
+  bucketName: v1-bucket
+`), nil)
+	require.NoError(t, err)
+
+	forked, err := out.Fork(map[string]string{"backup.storage.rootPath": "forked-root"})
+	require.NoError(t, err)
+
+	assert.Equal(t, "forked-root", forked.Backup.Storage.RootPath.Val)
+	assert.Equal(t, param.SourceOverride, forked.Backup.Storage.RootPath.Used.Kind)
+
+	assert.Equal(t, "milvus-proxy", forked.Milvus.Grpc.Address.Val)
+	assert.Equal(t, param.SourceV1ConfigFile, forked.Milvus.Grpc.Address.Used.Kind)
+	assert.Equal(t, "milvus.address", forked.Milvus.Grpc.Address.Used.Key)
+	assert.Equal(t, "v1-bucket", forked.Milvus.Storage.BucketName.Val)
 }
 
 // The v1 path validates what it produced, so a provider v1 never checked fails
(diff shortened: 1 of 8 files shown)


lentitude2tk (approved):
/LGTM

PR #1220, opened 2026-09-14
storage: treat an empty CopyObject ETag as a failed copy

huanghaoyuanhhy:
## Problem

S3 documents that a `CopyObject` which fails midway can still answer **200 OK with an `<Error>` document embedded in the body** (see [S3 CopyObject API](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CopyObject.html)). minio-go's `Client.CopyObject` decodes that body into a field-less `copyObjectResult` and reports success, leaving an **empty ETag for an object that was never copied**.

milvus-backup's `copyObject` discards the returned `UploadInfo` and trusts the error, so such a copy is recorded as done. The object then goes missing until the batched prefix verify (added in #1052) catches it as a missing key — at which point the whole backup fails with `verify prefix N objects missing`.

Observed in production on a first backup of a large collection: 130/130 segments logged "done" with zero warnings, yet 5 binlogs were absent from the backup bucket; the missing 5 were exactly the oldest/largest import files, and a full retry succeeded. GC was ruled out (GC was paused for the whole window and the files were marked valid by the leftover scan days earlier).

Notably, minio-go already handles the 200-with-embedded-error case for `CompleteMultipartUpload` and `RemoveObjects` via `expect200OKWithError`, but not for the single-part `CopyObject` path — so copies under the multipart threshold (default 500 MiB) are exposed.

## Fix

Treat an empty `UploadInfo.ETag` from `Client.CopyObject` as a failure so the existing `retry.Do` re-attempts the copy instead of silently dropping the object. This works on any minio-go version and does not depend on an upstream SDK fix.

## Tests

- `TestCopyObjectEmbeddedError`: reproduces the exact 200 + embedded `<Error>` response via a stubbed transport; fails before the fix (`An error is expected but got nil`) and passes after (the retry re-issues the copy and eventually surfaces an error).
- `TestCopyObjectValidETag`: pins the happy path so the empty-ETag guard does not misfire on a genuine copy.

Full `internal/storage` package passes; `go vet` clean.


lentitude2tk (approved):
/LGTM

PR #1221, opened 2026-09-14
ci: pull minio images from quay.io after docker hub takedown

huanghaoyuanhhy:
/kind improvement

Docker Hub deleted the `minio/minio` repository, so every compose-based CI job fails at `docker compose up` with `pull access denied for minio/minio, repository does not exist`, and the helm-based jobs time out inside `helm install --wait` while the minio pod sits in ImagePullBackOff — the chart's vendored minio subchart defaults to `minio/minio:RELEASE.2021-02-14T04-01-33Z`, which is gone from Docker Hub too.

This switches the five compose files under `deployment/` to `quay.io/minio/minio:RELEASE.2024-05-28T17-19-04Z` — MinIO's official quay namespace, same publisher, same release tag, verified present and anonymously pullable — and pins `minio.image.repository`/`minio.image.tag` in the three helm values files so they stop inheriting the dead subchart default. The helm tag moves from RELEASE.2021-02-14 to RELEASE.2024-05-28 because the 2021 tag no longer exists in any registry; that version bump is the only behavior change.

Verified locally: `docker compose config --quiet` passes on all five compose files, `helm template` renders the quay image for all three values files (standalone deploys a Deployment, cluster a StatefulSet — both covered), anonymous `docker pull` succeeds, and the image answers `/minio/health/live` with 200 when run with the exact command and env the compose files use.

The first CI run on this branch also surfaced a second, unrelated flake: the api jobs' Build step (which runs `build_image.sh`, a cold `docker build` whose `go mod download` starts from an empty cache on every runner) measured ~3.5 min on the last green run against a 5-minute step cap, and four api jobs were killed by ordinary network variance. The second commit raises that one step's cap to 10 minutes.


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

lentitude2tk (approved):
/LGTM

PR #1222, opened 2026-09-14
create: add --for purpose presets (secondary, archive, clone)

### LINKED ISSUE #1170: [Enhancement]: add purpose presets to create (e.g. --for secondary) so backup options match the restore contract

huanghaoyuanhhy:
Fixes #1170

Adds a purpose preset to create: `milvus-backup create -n b1 --for secondary`, plus `--for archive` and `--for clone`. A preset forces the option combination the matching restore path depends on, so one flag produces a backup that can serve that path instead of one that only looks successful and fails at restore time (the #1167 failure mode).

This is phase one: the preset lives entirely in the CLI. The core task and the REST API know nothing about purposes, and nothing is recorded in the backup meta — a follow-up can deepen it if the flag proves useful.

| knob | secondary | archive | clone |
|---|---|---|---|
| backup_index_extra | forced true | caller's choice | caller's choice |
| rbac | forced true | forced true | forced true |
| format | forced binlog | forced binlog | caller's choice (auto by default) |
| strategy | forced bulk_flush | caller's choice | caller's choice |

`--for secondary` mirrors what the secondary CI job passes by hand (`create --backup_index_extra --format=binlog`), plus the rbac and bulk_flush knobs a secondary needs.

Rationale for the parts the issue left open:

- `secondary` forces `format=binlog`: the secondary DML path builds its imports from partition/segment meta, and a snapshot-format backup has no segments, so it would import nothing while reporting success. The force also covers explicit `--format=auto`, which resolves to snapshot on a 3.x server.
- `archive` forces `binlog` too, as the format with the broadest restore compatibility for a long-lived artifact; `clone` is consumed by a plain restore right away, so it keeps the caller's format choice.
- `rbac=true` on all three: a secondary must serve after failover, an archive should carry the whole instance, and a clone is a copy of the instance.
- `backup_index_extra` is forced only on `secondary` — the etcd-sourced index extra info and dynamic field schemas it gates are what a secondary restore replays — and stays the caller's choice on `archive` / `clone`. It needs the `milvus.etcd` settings in backup.yaml; with those present, the #1218 startup probe makes an unreachable etcd fail the backup in its first seconds instead of silently leaving it incomplete (#1013).
- `secondary` forces `strategy=bulk_flush` exactly: it records the flush point the secondary replication starts from, and a serial flush adds nothing a secondary needs.

Semantics:

- A preset always wins over conflicting flags: `--for secondary --format=snapshot` runs with `format=binlog`. A conflict is overridden rather than reported, because refusing it would push the burden of spelling the combination correctly back onto the caller — the very thing the preset exists to absorb.
- The expansion is printed at create time (`for=secondary rbac=true format=binlog strategy=bulk_flush index_extra=true`).
- The REST `POST /api/v1/create` takes no purpose, and the backup meta records none: `--for` only ever changes the option values of the CLI create that carries it.


lentitude2tk (approved):
/LGTM

PR #1223, opened 2026-09-17
docs: add milvus 3.0 to the compatibility matrix

huanghaoyuanhhy:
/kind improvement

Add Milvus 3.0 to the README compatibility matrix: as a backup source row and as a restore target column. A footnote marks that backup support starts from 3.0.1 (3.0.0 is not supported); restore to 3.0 has no version restriction.


lentitude2tk (approved):
/LGTM

PR #1224, opened 2026-09-17
docs: update the compatibility matrix for 3.0 and drop restore to 2.4

huanghaoyuanhhy:
/kind improvement

Updates the README compatibility matrix:

- Adds Milvus 3.0 as a backup source row, missed by #1223. Both backup and restore require 3.0.1 or later; 3.0.0 is not supported in either direction.
- Drops the restore-to-2.4 column, which is no longer supported; the restore floor is now 2.5.


lentitude2tk (approved):
/LGTM

PR #1225, opened 2026-09-17
ci: test backup and restore against rustfs storage

huanghaoyuanhhy:
/kind improvement

Adds a standalone RustFS CI workflow: one job, master Milvus only.

RustFS (rustfs/rustfs) is an S3-compatible object store, so the new `deployment/standalone/docker-compose-rustfs.yml` is the existing compose with the MinIO service swapped for RustFS — same 9000 port, MinIO's default `minioadmin` credential pair so `configs/backup.yaml` needs no changes, and the standalone's `MINIO_ADDRESS` pointed at it. The workflow (`.github/workflows/rustfs.yaml`) deploys master-latest Milvus on it and runs the usual CLI round trip: prepare data, backup create, restore, verify, delete. It is a separate workflow so the main Test workflow's MinIO jobs stay untouched.

Image is pinned to `rustfs/rustfs:1.0.0` (first stable release) in line with the pinned MinIO/etcd images.


lentitude2tk (approved):
/LGTM

PR #1229, opened 2026-09-18
mpath: add fuzz tests for log path parsing

huanghaoyuanhhy:
/kind improvement

Add Go native fuzz targets for the mpath log path parsers — the code backup runs on every key it lists from milvus storage, which is input this tool does not control.

FuzzParseLogPath feeds arbitrary keys to both parsers and pins that no key panics them and that at most one accepts a key; its seed corpus includes the int64 overflow input from the #1197 fix so that regression stays guarded. FuzzInsertLogPathRoundTrip and FuzzDeltaLogPathRoundTrip build paths with the dir helpers from fuzzed ids and require the parsers to return exactly those ids, with the negative-id branch pinning that ids the regex cannot represent are rejected instead of misparsed.

The seed corpora run inside the normal `go test ./...`, so CI cost is unchanged. To actually search for failures: `go test ./internal/storage/mpath/ -fuzz=FuzzParseLogPath -fuzztime=30s`.


wenhuiZilliz (approved):

wenhuiZilliz (commented):
/LGTM

lentitude2tk (approved):
/LGTM

PR #1230, opened 2026-09-18
ci: drop milvus 2.4 axis from test-backup-restore-api

huanghaoyuanhhy:
Restore to Milvus 2.4 is no longer supported (the compatibility floor is now 2.5), but the test-backup-restore-api job still runs its suite against a 2.4-latest cluster. That job backs up and restores within the same cluster, so on 2.4 the restore half pins a capability users can no longer use, and a failure there is unactionable.

This drops the 2.4-latest axis; the job now runs only against master-latest. Backup-side coverage for 2.4 sources is untouched: the cross-version job in main.yaml still backs up from v2.4.23 and restores to supported targets.

/kind improvement


lentitude2tk (approved):
/LGTM

lentitude2tk (approved):
/LGTM

PR #1234, opened 2026-09-20
test: merge duplicate InferMapperRuleType tests into filter package

huanghaoyuanhhy:
/kind improvement

TestInferRuleType in core/server/restore_test.go tested filter.InferMapperRuleType with the same four positive cases as TestInferMapperRuleType in internal/filter/filter_test.go. It was a leftover of #917: the unify refactor moved the function into the filter package but only re-pointed the server test at the new location instead of moving it.

The two tests had disjoint error cases (two each), now merged into TestInferMapperRuleType; the server-side copy is deleted.


lentitude2tk (approved):
/LGTM

PR #1235, opened 2026-09-20
filter: drop unused AllowDBs and AllowNames batch methods

huanghaoyuanhhy:
/kind improvement

AllowDBs and AllowNames lost their last call sites in def9eb4 (#934), which replaced the list-everything-then-batch-filter flow in backup task with per-entry listing (listFilteredDBAndNSS). The singular AllowDB/AllowName remain in use by restore task. Remove the dead methods and their tests.


lentitude2tk (approved):
/LGTM

PR #1236, opened 2026-09-20
test: drop tautological strategy/format parse tests

huanghaoyuanhhy:
/kind improvement

Remove four tautological tests in `core/backup/func_test.go` that assert code against itself:
- `TestSupportStrategy` / `TestSupportFormat` assert `len(SupportStrategy()) == len(_strategyMap)`, i.e. `len(lo.Keys(m)) == len(m)`, which holds for any map.
- The `Valid` sub-tests of `TestParseStrategy` / `TestParseFormat` iterate the map's own keys and assert `ParseStrategy(key) == _strategyMap[key]`, i.e. `x == x`.

Both sides of each assertion move with the map, so no real behavior is pinned. Kept: `Empty` (defaults), `Invalid` (garbage input errors), and `SnapshotIsNotAStrategy` (snapshot is a format, error points at `--format=snapshot`). Real strategy/format keys stay pinned by `cmd/flags/removed_test.go` and `task_test.go`, and `cmd/create` derives its help text from the same maps, so no contract drift is possible.


lentitude2tk (approved):
/LGTM

PR #1237, opened 2026-09-20
storage: make prefix iterator construction infallible and lazy

huanghaoyuanhhy:
/kind improvement

## What

`Client.ListPrefix(ctx, prefix, recursive) (ObjectIterator, error)` becomes `Client.NewObjectIter(ctx, prefix, recursive) ObjectIterator`. The constructor can no longer fail: it always returns a usable iterator, and listing errors surface through `ObjectIterator.Next`, the same channel every caller already handles. The "iterator may be nil on error" footgun is gone by construction instead of by convention.

## Why the error return was safe to drop

Of the four implementations, only LocalClient could ever fail at construction (its eager stat/walk); the cloud implementations just wrap a pager or a channel that starts listing lazily and always returned a nil error. So the error return carried real information for exactly one implementation — this PR moves Local's listing behind the first `Next` call instead, which also unifies when failures happen across backends. Failure timing is the only behavior change: the same errors, one call later.

## Details

- The local iterator lists lazily: the stat/walk moves from construction to the first `Next`, tracked by a `listed` flag. A nonexistent prefix still lists as empty. `TestLocalClient_NewObjectIter/StatErrorSurfacesThroughNext` pins that a stat error surfaces through `Next`.
- The #1233 storage preflight in `app/create.go` simplifies accordingly: no more `if err == nil` nesting around the iterator, since construction cannot fail.
- Constructor-failure test cases (the `ListError` subtests) lose their premise and become iterator-error cases covering the same error-wrap branches. `client_mock.go` is regenerated; no other generated file is touched.
- `ObjectIterator` itself is unchanged, including the `Close` obligation. This is the stepping stone toward a range-over-func (`iter.Seq2`) listing API, which would retire `Close` entirely; kept as a separate PR so this diff stays mechanical.


lentitude2tk (approved):
/LGTM

PR #1238, opened 2026-09-20
storage: return prefix listing as iter.Seq2

huanghaoyuanhhy:
Follow-up to #1237, the second step of the prefix-iterator API reshape.

`Client.NewObjectIter` returned a pull-style `ObjectIterator` (`Next` + `Close`): every caller had to remember `Close`, and every implementation needed a comment explaining what `Close` was responsible for (minio: cancel the listing goroutine and drain its channel). This PR replaces the interface with `iter.Seq2[ObjectAttr, error]`, so the listing is driven by the range loop itself:

```go
for attr, err := range cli.NewObjectIter(ctx, prefix, true) {
    if err != nil {
        return err
    }
    // use attr
}
```

What this buys:

- **No Close to forget.** Leaving the loop early — `break`, `return`, or `panic` — stops the listing and releases its resources; the cleanup that `Close` used to own now runs as the sequence's defer on every exit path. The goroutine-leak class of bug the Close contract guarded against is gone by construction.
- **Consumer loops collapse to one statement.** `ListPrefixFlat`, `Exist`, `DeletePrefix`, `CopyPrefixTask`, `VerifyPrefixTask` and the create source-storage preflight all drop their `for { Next; if !ok break }` scaffolding. `Exist` and the preflight now read as "take the first yielded object and return".
- **Independent listings.** Each range over the returned sequence runs a fresh listing; all state lives inside the closure body, nothing is shared across ranges.

Behavior is unchanged from what #1237 established: construction stays lazy (no request until the first iteration), listing errors surface as the last yielded value, exhaustion simply ends the range. The minio implementation keeps its cancel-and-drain — the early-stop test now observes the cancellation reaching the in-flight page request directly, a tighter assertion than the old post-Close exhaustion check. Local keeps lazy stat/walk; azure's pager loop keeps the empty-page and first-`More()` semantics; gcp wraps the GCS iterator unchanged.

`NewMockObjectIterator` keeps its name and becomes a sequence constructor, so the cross-package test stubs (app, core, meta) barely change. `client_mock.go` regenerated with `make gen-mock`.

/kind improvement


lentitude2tk (approved):
/LGTM

PR #1240, opened 2026-09-22
ci: pin 2.6-latest image to known-good daily

huanghaoyuanhhy:
/kind improvement

Pin the `2.6-latest` short tag to `milvusdb/milvus:2.6-20260920-74a6cd84` in `scripts/get_image_tag_by_short_name.py`, the single resolution point used by every `2.6-latest` call site in `main.yaml` and `nightly.yaml`.

Why: daily builds from `2.6-20260920-b1aa37dc` on (milvus-io/milvus#53540 backport) kill standalone during the L2 restore test — the process dies with no panic in its log (SIGKILL signature) and 19530 starts refusing connections. `2.6-latest, L2` has failed on every run since, including on main's scheduled runs, while L0/L1 and master-latest L2 pass on the same image. `74a6cd84` is the last daily before the bad commit; upstream has no fix or revert yet, so the pin stays until one ships.

Also record container exit state in `scripts/export_log_docker.sh`: `docker inspect` exit code / OOM flag / finished-at plus `docker compose ps -a` go into the artifact next to the logs, so the next silent death carries its exit code instead of just a truncated log.


lentitude2tk (approved):
/LGTM

PR #1242, opened 2026-09-22
backup: gate the snapshot format on same-service storage

### LINKED ISSUE #1241: backup: refuse the snapshot format when the backup storage is not on the milvus storage's service
Since #1137, the default auto format resolves to snapshot on Milvus 3.0 servers. The snapshot copy is executed by Milvus as a server-side copy inside one storage service, so when backup.storage names a different provider than milvus.storage — or a second self-hosted MinIO endpoint — the export jobs fail deep inside Milvus with errors that do not point at the configuration.

Before that default changed, such cross-provider setups were served by the binlog format, which streams data through milvus-backup and moves it between any two supported backends. On 3.0 the binlog format is no longer an option, so this configuration has no working backup format at all, and nothing says so.

milvus-backup should detect the mismatch before any collection work starts and refuse the backup (and the restore, whose copy runs the same way in the opposite direction) with an error naming both storage configs. Same-service cases must keep working: another bucket, account, or region of the same cloud provider, including cross-account Azure copies with a source SAS.

huanghaoyuanhhy:
## What

Refuse snapshot-format backups and restores when the milvus storage and the backup storage are not the same storage service, instead of letting the server-side copy fail deep inside Milvus.

## How

- internal/storage gains SnapshotSameService: same provider family required (s3 and aws are one family, gcp and gcpnative are not); a cloud provider's service spans its accounts and regions; a self-hosted MinIO is its endpoint as Milvus reaches it, so two endpoints are two services. Azure cross-account copies keep working through the existing source SAS path.
- backup: both auto and --format=snapshot fail with an error naming both storage configs when the two are not the same service. There is no binlog fallback to offer on a 3.0 server, so auto errors rather than silently downgrading.
- restore: a snapshot-format backup held on another service is refused before any job is submitted.
- README documents the snapshot format's same-service requirement.

Fixes #1241


lentitude2tk (approved):
/LGTM

PR #1244, opened 2026-09-22
restore: allow shard_num override equal to bundle shards on snapshot

huanghaoyuanhhy:
/kind improvement

Snapshot restores refused every per-collection `shard_num` override, including one equal to the bundle's shard count, where the restore already produces exactly the collection the override asks for.

This moves the check out of `checkSnapshotSupport` into the snapshot collection task, which knows the bundle's real shard count. An override equal to the bundle's count now proceeds, and an override naming a different count is still refused, because Milvus creates the collection from the bundle and shards are not alterable afterwards. The request-level `max_shard_num` gate is unchanged.


lentitude2tk (approved):
/LGTM

